PR #109, opened 2019-06-20
Feature/task 76 have clear indication of super user

### LINKED ISSUE #76: Have clear indication if logged in as a SuperUser

andriikovalov-dlr:
If logged in as a super user, set a custom eclipse theme, where toolbar and status bar have red background

Closes #76


andriikovalov-dlr:
@PhilMFischer Yes, but I only toolbar and status bar. We can customize it more.
super_user_theme

Other controls also look slightly different. Apparently, the default "base" theme for a css file is different from the default eclipse theme. To make it look identical we could also create os-specific versions referencing os-specific css files for default theme, but I'm not sure it's worth the effort.

PR #130, opened 2019-07-08
Added waitForEditor condition for SWTBotTest cases

### LINKED ISSUE #129: Check for editor opening in SWTBotTest cases
Add a wait condition in the openEditor method of the ASWBotTestCase class.
Right now the method only has a time based waiting condition, but not a logic based waiting condition. This can lead to unstability of the opening of the editor lags.

SaMuellerDLR:
Close #129

### DIFF
--- a/de.dlr.sc.virsat.swtbot.test/src/de/dlr/sc/virsat/swtbot/test/ASwtBotTestCase.java
+++ b/de.dlr.sc.virsat.swtbot.test/src/de/dlr/sc/virsat/swtbot/test/ASwtBotTestCase.java
@@ -9,6 +9,7 @@
  *******************************************************************************/
 package de.dlr.sc.virsat.swtbot.test;
 
+import static org.eclipse.swtbot.eclipse.finder.matchers.WidgetMatcherFactory.withPartName;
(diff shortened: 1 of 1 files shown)


de.dlr.sc.virsat.swtbot.test/src/de/dlr/sc/virsat/swtbot/test/ASwtBotTestCase.java
```diff
@@ -145,11 +148,25 @@ protected void save() {
...
+	
+	/**
+	 * Waits for the editor of the passed item to open.
+	 * The editor is identified using the name of the item,
+	 * so it cant distinguish between editors of items with the same name!
+	 * @param item the item whose editor we are waiting for
+	 */
+	protected void waitForEditor(SWTBotTreeItem item) {
+		String label = item.getText();
+		Matcher<IEditorReference> matcher = withPartName(StringStartsWith.startsWith(label + " -> "));
```
PhilMFischer: Does it make sense to create the editor name here by hand. Shouldn't it be derived from a method actually creating the name in the editor. SO in case the pattern changes, the test case won't fail?

Issue #220, opened 2019-08-16
Clean up mechanical export products


PhilMFischer:
Currently we export the shape and even STL file references in the products part of the mechanical export.

The issue is, that a product assembles parts. Thus a product imprts several parts and arranges them in their position and orientation.

The Configuration tree in Virtual Satellite represents this assembly of products and even sub products. Usually a sub product / sub assembly is marked by a ConffigurationElement with NONE shape. In case there is a shape selected, it should be export the part as well and assemble it into the product in the origin position. The part is then exported from the ElementDefinition or the ElementConfiguration represents its own part.

Additionaly:
- [ ] if a visualization is set to NONE, it should not be exported as part
- [ ] if a visualization within a tree is set to none it fulfils the purpose of transformation. This emans for the configuration tree, that it has to be exportet, but it is not referencing another part.

PR #222, opened 2019-08-16
WIP: Cleaned up properties in Product structure of json - (Task #220)

ghost:
- Not exporting shape anymore
- Not exporting stl anymore
- renamed properties to be more consistent
- added testcase to check for sub assemblies and their part
representatives in the json export
- adjusted the imports accordingly

---
Task #220: Clean up mechanical export products


franzTobiasDLR (dismissed):

andriikovalov-dlr (commented):
This kind of makes sense to avoid ambiguity in where shape is defined. But what if we override shape in a Configuration (e.g. add an stl), then with this approach it will be ignored in the export, then if there are changes to the shape on CAD side they will be reimported into Definition, but Configuration won't change (since it overrides).
If this is intended, I'll approve it.

andriikovalov-dlr (dismissed):

SaMuellerDLR:
Created new PR to replace this one https://github.com/virtualsatellite/VirtualSatellite4-Core/pull/322

PR #248, opened 2019-08-30
Migrating TravisCI to xenial and beyond

ghost:
- Changed travis script to got to xenial
- Added new apt dependencies for loading zmq and vtk 
- Refactored external lib plugins for linux now using external libraries rather then shiopped ones
- added new script to prepare the environment
- Still with openJDK 8, therefore not yet on  bionic which does not support 8 anymore.
- Removed all so files which were sled dependant

:warning: Be careful with merging, all other depending projects have to be prepared, otherwise they will fail to build on TravisCI due to missing environment. :warning:


andriikovalov-dlr (approved):

de.dlr.sc.virsat.swtbot.test/src/de/dlr/sc/virsat/swtbot/test/ASwtBotTestCase.java
```diff
@@ -52,6 +55,19 @@
  */
 public class ASwtBotTestCase {
 
+	@Rule
+	public TestRule testStatusWatcher = new TestWatcher() {
+		@Override
+		protected void starting(Description description) {
+			System.err.println("SWTBotTestCase: " + description.getMethodName() + "...started");
```
SaMuellerDLR: Any reason to output it in System.err instead of System.out?

SaMuellerDLR (changes requested):

de.dlr.sc.virsat.swtbot.test/src/de/dlr/sc/virsat/swtbot/test/ASwtBotTestCase.java
PhilMFischer: The basic idea was to see if it is possible to get direct out put into the TravisCI log rather then the redirected EclipseLog which is redirected into the file. Since it did not work i changed it to standard out now.

SaMuellerDLR (approved):
Approved

PR #261, opened 2019-09-05
Load referenced DMF resource and resolve proxy - (Task #259)

### LINKED ISSUE #259: DMF resource cannot handle references to DMF models of other resources
DMF resource and its save command cannot handle DMF references to model elements of other resources. Command and resource do not resolve proxies and do not load referenced resources... Which leads to NPE and IndexOutOfBoundExceptions. 

To reproduce create a tasking environment on one structural element and a task definition on another... Then instantiate the task definition in the task diagram. Saving the diagram results in several errors.

franzTobiasDLR:
Load resources which are referenced by DMF model elements and resolve their proxies

Closes #259


PhilMFischer:
@franzTobiasDLR Are you talking about referencing objects in one DMF into another DMF model? Is that what you addressin #259?

franzTobiasDLR:
> 
> 
> @franzTobiasDLR Are you talking about referencing objects in one DMF into another DMF model? Is that what you addressin #259?

Yes

bash/maven_build.sh
```diff
@@ -45,7 +45,10 @@ printUsage() {
 
 checkforMavenProblems() {
 	echo "Check for Maven Problems on Product:"
-	(grep -n "\[\(WARN\|WARNING\|ERROR\)\]" maven.log | grep -v "\[WARNING\] Checksum validation failed" || exit 0  && exit 1;)
```
PhilMFischer: This one should have been handled by the PR #264. Please check that there is no LF/CR issue here. There should be no difference to Development.

de.dlr.sc.virsat.model.extension.tests.edit.test/src/de/dlr/sc/virsat/model/dvlm/dmf/DmfResourceTest.java
```diff
@@ -659,8 +660,85 @@ protected void doExecute() {
 		TestCategoryReference beanTestCategoryAssignmentReference = new TestCategoryReference(referencingCa);
 		
 		assertEquals("Reference is set correctly", beanTestCategoryAssignment, beanTestCategoryAssignmentReference.getTestRefCategory());
+		
+		resSeiOther.unload();
+		resSei.unload();
+		dmfCategoryAssignmentReference.setName("vfdv");
```
PhilMFischer: Why setting these strange names?

de.dlr.sc.virsat.model.extension.tests.edit.test/src/de/dlr/sc/virsat/model/dvlm/dmf/DmfResourceTest.java
```diff
@@ -659,8 +660,85 @@ protected void doExecute() {
...
+
+		dObjectContainerOther.getObjects().add(dmfCategoryAssignment);
+		dmfResourceOther.save(Collections.EMPTY_MAP);
+		
+		DObjectContainer dObjectContainer = (DObjectContainer) dmfResource.getContents().get(0);
+		dmfCategoryAssignmentReference.setTestRefCategory(dmfCategoryAssignment);
+		dObjectContainer.getObjects().add(dmfCategoryAssignmentReference);
+		dmfResource.save(Collections.EMPTY_MAP);
+		
+		//Unload original DVLM resources
```
PhilMFischer: Please add explanation why to undload the DVLMs. What is the expected bahvior. And what should happen in the backgrund?

de.dlr.sc.virsat.model.extension.tests.edit.test/src/de/dlr/sc/virsat/model/dvlm/dmf/DmfResourceTest.java
```diff
@@ -659,8 +660,85 @@ protected void doExecute() {
...
+		resSeiOther.unload();
+		resSei.unload();
+		
+		//Check that its still possible to save changes
+		final String changedNameReference = "NewNameReference";
+		final String changedNameReferenced = "NewNameReferenced";
+		dmfCategoryAssignmentReference.setName(changedNameReference);
```
PhilMFischer: Why just changing the name of the reference and not the actual reference?

de.dlr.sc.virsat.model.extension.tests.edit.test/src/de/dlr/sc/virsat/model/dvlm/dmf/DmfResourceTest.java
```diff
@@ -659,8 +660,85 @@ protected void doExecute() {
...
+		dmfCategoryAssignment.setName(changedNameReferenced);
+		dmfResource.save(Collections.EMPTY_MAP);
+		dmfResourceOther.save(Collections.EMPTY_MAP);
+
```
PhilMFischer: As far as i understood, by saving, the underlying DVLM should be opened and loaded before the changes are saved to it. Right? Thus, would it make sense, to check if the DVLM resources are wll unloaded before the save statement, and then check if the DVLM resources got correctly laoded?

de.dlr.sc.virsat.model.extension.tests.edit.test/src/de/dlr/sc/virsat/model/dvlm/dmf/DmfResourceTest.java
franzTobiasDLR: Oh that's left over from when I was debuging it, in the dedicated test case for it I use proper names

franzTobiasDLR: I will remove it from this test case.

de.dlr.sc.virsat.model.extension.tests.edit.test/src/de/dlr/sc/virsat/model/dvlm/dmf/DmfResourceTest.java
franzTobiasDLR: Checking that the resources are properly unloaded makes sense, yes. I add an assert statement for that. Checking if the save operation worked is done by checking if the changes appear in the DVLM resources. I don't think it is necessary to check if the DVLM resource are loaded afterwards as from a DMF perspective it does not matter. Important is that the changes have bean saved.

Issue #267, opened 2019-09-06
Set all files to windows style CR/LF


PhilMFischer:
Some files are set to linux. Changing them may result in saving themw ith different line endings. Changing all to windows will improve comapring files.

All files except to eb changed except:
- .gitignore
- .mailmap
- known_authors
- bash/*

PR #272, opened 2019-09-09
Changed all files to windows style CR/LF

ghost:
- Changed all files to be on windows CR/LF.
- Did not change scripts in bash folder since they are executed on linux.
- Did not change .mailmap and .gitignore and known_authors.txt file.
- Setup all projects to use windows LF as standard.
- Fix new requirements plugins as well


SaMuellerDLR (approved):
Approved!

franzTobiasDLR (approved):

PhilMFischer:
#267 can be forwarded to CEF now.

Issue #274, opened 2019-09-10
Requirement feature is not in p2 updatesite


franzTobiasDLR:
Requirement feature is not in p2 updatesite but required by the test plugin...

PR #275, opened 2019-09-10
Add requirements feature to p2 site and tmodel - (Task #274)

franzTobiasDLR:
Add the requirements feature to p2 project and tmodel files because it is required by the test plugin and thus needs to be deployed...


PhilMFischer:
Closes #274

PR #319, opened 2019-10-25
Added Try Catch to avoid NPE when executing on linux

### LINKED ISSUE #318: NPE when executing under linux with incorrect VTK and ZMQ
Starting Virtual Satellite under Linux requires that VTK and ZMQ are setup correctly on the Linux computer. System variables have to be set for loading the libraries. If things are incorrect here, Virtual Satellite starts the Visualization Plugins which try to run a SceneGraphServer. This one requires VTK and ZMQ and will fail loading the classes. This disturbs usability.

ghost:
The Visualization plugins try to directly start VTK and ZMQ from their activators. This causes annoying NPEs in case ZMQ and VTK are not well setup on the linux system. This PR introduces some try catch to avoid the crashes and provide reasonable log entries instead.

Actively opening a VTK Viewer in case VTK is not correctly setup still creates NPEs.

Closes #318

### DIFF
--- a/de.dlr.sc.virsat.model.extension.visualisation.ui/src/de/dlr/sc/virsat/model/extension/visualisation/ui/Activator.java
+++ b/de.dlr.sc.virsat.model.extension.visualisation.ui/src/de/dlr/sc/virsat/model/extension/visualisation/ui/Activator.java
@@ -43,7 +43,7 @@ public class Activator extends AbstractUIPlugin {
 	private CommunicationServer geometryFileServer;
 	private ResourceReloadListener resourceReloadListener;
 
-	private boolean serverstarted = false;
+	private boolean serverStarted = false;
 	
 	/**
 	 * The constructor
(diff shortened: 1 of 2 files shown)


ghost (approved):
Approved

de.dlr.sc.virsat.model.extension.visualisation.ui/src/de/dlr/sc/virsat/model/extension/visualisation/ui/Activator.java
```diff
@@ -42,6 +43,8 @@
 	private CommunicationServer geometryFileServer;
 	private ResourceReloadListener resourceReloadListener;
 
+	private boolean serverstarted = false;
```
andriikovalov-dlr: Let's make it camel case like serverStrarted

andriikovalov-dlr (changes requested):

andriikovalov-dlr (approved):

PR #352, opened 2019-11-13
Feature/task 351 proxy in dmf save command

### LINKED ISSUE #351: DmfResourceSaveCommand contains proxy to SEI
In some rare cases, the DMFReosurce creates a SaveCommand from a SEI-proxy. The command should make sure, that the containing SEI is properly loaded.

Hard to reproduce as it only happens after a DMFResource is opend in a editor for some time and not edited.

franzTobiasDLR:
Resolve proxies in DMFSaveCommand...

When first using the SEI in the command, it checks if it is a proxy... if so it resolves it.

Closes #351


SaMuellerDLR:
@franzTobiasDLR Is there some whitespace change going on? Im seeing a lot of line changes,

franzTobiasDLR:
That's why I didn't request review yet... It's somehow strange, my local files before and after my change both have Windows file endings...

PR #350, opened 2019-11-11
Created generator class for validator tests

Anna-Lena-DLR:
- validator tests are generated when creating a new concept
- validate() is tested

PR #354, opened 2019-11-18
Give validator concept specific name

### LINKED ISSUE #100: Concept-specific name for Validator class
Our generated validator is called StructuralElementInstanceValidator Therefore, we have a class with this name in every concept, which might be a bit confusing. It would be better to have something like ScosaSeiValidator

Anna-Lena-DLR:
- When a new concept is created, the generated validator gets a concept-specific name "conceptName"Validator.
- Tests are adjusted to that.
- existing AStructuralElementInstanceValidators are marked as depricated
- plugin.xml updated

Also the generated validator tests get a concept-specific name.
closes #100


franzTobiasDLR:
Should be merged after #350

de.dlr.sc.virsat.model.concept.test/resources/expectedOutputFilesForGenerators/AStructuralElementInstanceValidator.java
```diff
@@ -25,11 +25,10 @@
  * 
  * 
  */	
-public abstract class AStructuralElementInstanceValidator implements IStructuralElementInstanceValidator {
+public abstract class AtestConceptValidator implements IStructuralElementInstanceValidator {
```
PhilMFischer: The name of the class does not fit to the file name. This is unusual for java. Here it is a question of the test case, most likely the testcase is still called a StructuralElementInstance… Please check and adjust.

Just in case, the test output can also be generated and does not need to be adjusted by hand.

de.dlr.sc.virsat.model.concept.test/resources/expectedOutputFilesForGenerators/AStructuralElementInstanceValidatorTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Auto Generated Abstract Generator Gap Class
+ * 
+ * Don't Manually modify this class
+ * 
+ * 
+ * 
+ */	
+public abstract class AtestConceptValidatorTest {
```
PhilMFischer: Same as above. class and file names do not match.

de.dlr.sc.virsat.model.concept.test/src/de/dlr/sc/virsat/model/concept/generator/tests/GenerateValidatorTestsTest.xtend
```diff
@@ -0,0 +1,76 @@
...
+	def void testCreateForCategoryTests() {
+		concept = '''
+			Concept «TEST_CONCEPT_NAME» {
+			}
+			'''.parse
+		
+		val concreteClassContents = generateValidatorTests.createConcreteClass(concept, concept)
+		val abstractClassContents = generateValidatorTests.createAbstractClass(concept, concept)
+		
+		GeneratorJunitAssert.assertEqualContent(concreteClassContents, "/resources/expectedOutputFilesForGenerators/StructuralElementInstanceValidatorTest.java")
```
PhilMFischer: The names of the expected output files should also be adjusted here.

de.dlr.sc.virsat.model.concept.test/src/de/dlr/sc/virsat/model/concept/generator/validator/GenerateValidatorTest.xtend
```diff
@@ -45,7 +45,7 @@ class GenerateValidatorTest {
     def void testCreateConcreteClassFileName() {
     	val fileName = validatorGenerator.createConcreteClassFileName(concept, concept)
     	
-    	val expectedFileName = testConceptName + "/validator/StructuralElementInstanceValidator.java"
+    	val expectedFileName = testConceptName + "/validator/testConceptValidator.java"
```
PhilMFischer: By convention class names should start with a capital letter "TestConceptValidator.java" same as the class.

de.dlr.sc.virsat.model.concept/src/de/dlr/sc/virsat/model/concept/generator/tests/GenerateAllTests.xtend
```diff
@@ -48,7 +48,7 @@ class GenerateAllTests extends AGeneratorGapGenerator<EObject> {
 	public static val PACKAGE_FOLDER = "test";
 	
 	override serializeModel(Concept concept, IFileSystemAccess fsa) {
-				
+
```
PhilMFischer: PLease check if there is an empty line. Most likely we don't need it here.

de.dlr.sc.virsat.model.concept/src/de/dlr/sc/virsat/model/concept/generator/tests/GenerateAllTests.xtend
```diff
@@ -189,11 +189,13 @@ class GenerateAllTests extends AGeneratorGapGenerator<EObject> {
 			listOfTests.add(it + "Test")
 		]
 		
+		importManager.register(concept.fullQualifiedName + ".validator." + concept.name.replace("de.dlr.sc.virsat.model.extension.", "") + "Validator" + "Test")
```
PhilMFischer: What is this about? Maybe a comment in the code could help ehre a bit. I am confused by the replace statement. Nevertheless i guess there is a good reason for it.

de.dlr.sc.virsat.model.concept/src/de/dlr/sc/virsat/model/concept/generator/validator/GenerateValidator.xtend
```diff
@@ -24,20 +24,20 @@ class GenerateValidator extends AGeneratorGapGenerator<EObject> {
 		concept.name + "." + PACKAGE_FOLDER;
 	}
 	
-	static def getConcreteClassName(EObject typeDefinition) {
-		"StructuralElementInstanceValidator";
+	static def getConcreteClassName(Concept concept) {
+		concept.name.replace("de.dlr.sc.virsat.model.extension.", "") + "Validator";
```
PhilMFischer: Two things here:
1. There are now several places where we need to extract the concept name from the FQN. This should be refactored into some utility class or similar. With xtend it should allow a simple call here.
2. The current code will create invalid class names since it does not obey that the first letter of a class should be written in upper case. 

-> Please consider to externalize this code and also provide a test for correctly extracting the class name.

de.dlr.sc.virsat.model.concept/src/de/dlr/sc/virsat/model/concept/generator/validator/GenerateValidator.xtend
```diff
@@ -24,20 +24,29 @@ class GenerateValidator extends AGeneratorGapGenerator<EObject> {
 		concept.name + "." + PACKAGE_FOLDER;
 	}
 	
-	static def getConcreteClassName(EObject typeDefinition) {
-		"StructuralElementInstanceValidator";
+	static def getValidatorName(Concept concept) {
```
PhilMFischer: Please add some comment description to this method.

franzTobiasDLR:
> 
> 
> Looks fine to me.
> 
> @franzTobiasDLR What about the backwards compatibility as stated in the review before?
> 
> Cheers

Oh, I didn't see the comment or me being mentioned... I will look into it.

franzTobiasDLR:
Looks like the new validators are not yet registered in the concepts' projects Plugin.xml. I think we have to also update the Plugin.xml-Generator for Concepts. ;)

However, I think we can't simply change the registration. This would disable existing validators in projects that have a custom validator implementation already. A proper way to solve this would probably be to add a new registration in the Plugin.xml (and leave the old one if exists). Old validator classes should then be marked as deprecated and trigger a warning. 

Let's maybe discuss a solution for that next week :)

Issue #324, opened 2019-11-07
VirSat Exporters and Importers are not in the same group


andriikovalov-dlr:
![image](https://user-images.githubusercontent.com/48556822/68373727-bd626400-0143-11ea-992b-03eb1a844d16.png)

![image](https://user-images.githubusercontent.com/48556822/68373789-dff47d00-0143-11ea-977c-7f45580aa8a2.png)

PR #371, opened 2019-11-25
Updated categories of VirSat Importers/Exporters - Task (#324)

SaMuellerDLR:
- Updated the categories of the Excel and the Funcelectrical exporters / importers
- Improved name consistencies


de.dlr.sc.virsat.model.extension.funcelectrical.ui/plugin.xml
```diff
@@ -538,16 +538,16 @@
...
+        id="de.dlr.sc.virsat.model.extension.funcelectrical.ui.exportWizard"
+        name="Functional Electrical Architecture to HTML Export Wizard">
 		</wizard>
-		<category
-			id="de.dlr.sc.virsat.excel.ui"
-			name="Excel Export">
-		</category>
+      <category
+            id="de.dlr.virsat.global.category.export.virsat"
+            name="VirSat">
```
PhilMFischer: Just a remark. Maybe we should call the category "Virtual Satellite". But this could also be an individual ticket.

PR #395, opened 2019-12-10
Improved logging for null objects in resource - (Task #394)

### LINKED ISSUE #394: HotFix: Improve Logging of Model Resource
Somehow a null objects gets into the content of the EMF model resources.

The reason is not clear yet. 

Therefore improved logging is needed that:
- [x] gives feedback when a null object is added
- [x] gives a detailed stack trace from where this operation is triggered
- [x] tells which resource is actually affected

ghost:
- Added an adapter to the resources to listen for changes that may
introduce a null. The adapter will ouput a stack trace to the log and
mention the resource. After that the diagnostics will be updated to
notify the editor.
- Added some code to check the resource on null objects when performing
the updatediagnostics. A diagnostics error will be created in that case.
This diagnostics are used to inform the user by the editors.

---
Task #394: HotFix: Iprove Logging of Model Resource

Closes #394


SaMuellerDLR (approved):
Approved!

andriikovalov-dlr (approved):
Approving for a hotfix, but for merging to development we would need to polish it more

Issue #392, opened 2019-12-10
HotFix: Calculations stop silently on error


PhilMFischer:
The calculation engine stops silently on errors.

This needs to be improved to at least improve logging:

- [x] Improve logging
- [x] Check if UI feedback is possible as well.

Related to #387

andriikovalov-dlr:
We will only write errors to log. UI feedback would require substantial changes to `VirSatProblemMarkerHelper` because it only supports adding markers to `EObject`, but not EMF `Resource`.

PR #403, opened 2019-12-12
Error reporting in equation builder on errors in resources - (Task #392)

andriikovalov-dlr:
In equation builder if there are errors in ResourceSet, write them to log and show them to the user.

This has to be also merged to development.

### DIFF
--- a/de.dlr.sc.virsat.model.calculation/src/de/dlr/sc/virsat/model/calculation/compute/IncrementalEquationBuilder.java
+++ b/de.dlr.sc.virsat.model.calculation/src/de/dlr/sc/virsat/model/calculation/compute/IncrementalEquationBuilder.java
@@ -33,7 +33,6 @@ import org.eclipse.emf.common.command.Command;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.emf.ecore.util.EcoreUtil;
-import org.eclipse.ui.statushandlers.StatusManager;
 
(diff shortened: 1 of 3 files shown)


de.dlr.sc.virsat.model.calculation/src/de/dlr/sc/virsat/model/calculation/compute/IncrementalEquationBuilder.java
```diff
@@ -348,8 +349,37 @@ public boolean visit(IResourceDelta delta) throws CoreException {
...
+	 * @param resource EMF resource with an error
+	 * @param error 
+	 * @return status for writing to log
+	 */
+	private Status getResourceErrorStatus(Resource resource, Resource.Diagnostic error) {
+		Status status;
+		Throwable exception = error instanceof Throwable ? (Throwable) error : null;
+		status = new Status(Status.ERROR, Activator.getPluginId(),
+				String.format("IncrementalEquationBuilder: error in resource %s: %s", resource.toString(),
+						error.getMessage()),
```
PhilMFischer: May create a NPE in case that error is not of type throwable.

andriikovalov-dlr: If error is not throwable, exception will be null, which is accepted in `Status` constructor:

> **exception** a low-level exception, or `null` if not applicable

(from Status javadoc)

PR #406, opened 2019-12-12
Using locking mechanism to protect access of commando execution

SaMuellerDLR:
- Introduced a lock in the VirSatWorkspaceCommandStack to control who is allowed to execute / undo / redo.
- Using the locks atomic "tryLock" mechanism in VirSatSaveJob and the builders to prevent protential deadlocks
- Refactored the VirSatTransactionalEditingDomain handling of recently changed resources, to clear them after processing is done.


SaMuellerDLR:
Should be broken down.. too much critical change.

PR #435, opened 2020-01-22
Feature/task 434 correct codacy issues

Anna-Lena-DLR:
- Changed assertTrue to assertFalse
- Remove unnecessary assertTrue statements
- Remove empty test methods


de.dlr.sc.virsat.model.concept.test/src/de/dlr/sc/virsat/model/concept/generator/ConceptOutputConfigurationProviderTest.java
```diff
@@ -87,7 +87,6 @@ public void testGetOutputConfigurations() {
 					assertFalse("SetDerivedProperty is set correctly", outputConf.isSetDerivedProperty());
 					break;
 				default:
-					assertTrue("All names are correct", false);
 					break;
```
SaMuellerDLR: Dont we need a fail("All names are correct") here? Also I am wondering about the message. Is it correct?
Since its a fail, I would expected some negative statement.

PR #452, opened 2020-01-30
Added build specific test artifact deployment

### LINKED ISSUE #450: Add build specific artefact uplaod to build process
At the moment swtbot screen captures are uploaded to Sourceforge. A new build is overwriting these images.

What si required is a build specific zip which will additionally contain aretfacts such as the log outputs when tests are executed.

ghost:
- SWTBOT Artifacts are now zipped and uploaded to sourceforge
- zip files are individually named, thus they don't get overwritten
- Updated scripts so that maven script can be run locally on Linux computers
- Updated deploy script, so it only deploys in case of test artifacts exist
- Changed strategy on codecov reporting to try to send reports as early as possible. which means also on failed test cases. This may improve finding runtime issues on failing test cases due to race conditions.

Closes #450


bash/deploy_sourceforge.sh
```diff
@@ -33,8 +33,22 @@ printUsage() {
...
-	rsync -e ssh -avP $TRAVIS_BUILD_DIR/swtbot/  dlrscmns@frs.sourceforge.net:/home/frs/project/virtualsatellite/VirtualSatellite4-Core/swtbot/
+	if [ "$(ls -A ${TRAVIS_BUILD_DIR}/swtbot)" ]; then
+		echo "Test Artifacts: Collecting all reports in: ${TRAVIS_BUILD_DIR}/swtbot"
+		cp $TRAVIS_BUILD_DIR/de.dlr.sc.virsat.swtbot.test/target/surefire-reports/* $TRAVIS_BUILD_DIR/swtbot/
+		TEST_ARTEFACTS_ZIP=SwtBot_${TRAVIS_REPO_SLUG}_${TRAVIS_BRANCH}_${TRAVIS_JOB_NUMBER}_${TRAVIS_COMMIT}
+		echo "Test Artifacts: About to zip all artifacts into: ${TEST_ARTEFACTS_ZIP}"
+		TEST_ARTEFACTS_ZIP=$(sed -e 's/\//\_/g' <<< ${TEST_ARTEFACTS_ZIP})
+		echo "Test Artifacts: About to zip all artifacts into: ${TEST_ARTEFACTS_ZIP}"
+		TEST_ARTEFACTS_ZIP=$(sed -e 's/\./\_/g' <<< ${TEST_ARTEFACTS_ZIP}).zip
+		echo "Test Artifacts: About to zip all artifacts into: ${TEST_ARTEFACTS_ZIP}"
```
andriikovalov-dlr: Let's remove intermediate debug echoes

bash/deploy_sourceforge.sh
```diff
@@ -33,8 +33,22 @@ printUsage() {
...
-	cp $TRAVIS_BUILD_DIR/de.dlr.sc.virsat.swtbot.test/target/surefire-reports/* $TRAVIS_BUILD_DIR/swtbot/
-	rsync -e ssh -avP $TRAVIS_BUILD_DIR/swtbot/  dlrscmns@frs.sourceforge.net:/home/frs/project/virtualsatellite/VirtualSatellite4-Core/swtbot/
+	if [ "$(ls -A ${TRAVIS_BUILD_DIR}/swtbot)" ]; then
+		echo "Test Artifacts: Collecting all reports in: ${TRAVIS_BUILD_DIR}/swtbot"
+		cp $TRAVIS_BUILD_DIR/de.dlr.sc.virsat.swtbot.test/target/surefire-reports/* $TRAVIS_BUILD_DIR/swtbot/
+		TEST_ARTEFACTS_ZIP=SwtBot_${TRAVIS_REPO_SLUG}_${TRAVIS_BRANCH}_${TRAVIS_JOB_NUMBER}_${TRAVIS_COMMIT}
+		echo "Test Artifacts: About to zip all artifacts into: ${TEST_ARTEFACTS_ZIP}"
+		TEST_ARTEFACTS_ZIP=$(sed -e 's/\//\_/g' <<< ${TEST_ARTEFACTS_ZIP})
+		echo "Test Artifacts: About to zip all artifacts into: ${TEST_ARTEFACTS_ZIP}"
+		TEST_ARTEFACTS_ZIP=$(sed -e 's/\./\_/g' <<< ${TEST_ARTEFACTS_ZIP}).zip
```
andriikovalov-dlr: Can we maybe substitute both / and . in one sed command?

andriikovalov-dlr (changes requested):
I'm wondering if we are deploying all swtbot results then who is going to delete them?

bash/deploy_sourceforge.sh
PhilMFischer: Yes tried it. Have a look

bash/deploy_sourceforge.sh
PhilMFischer: Tried it as well. seems to work on a local Linux machine.

SaMuellerDLR (approved):
Approved!

andriikovalov-dlr (approved):

PR #479, opened 2020-02-19
Threading: Change Editing domain to lock workspace on tryRunExclsuive

### LINKED ISSUE #467: Multi-select not shown in VirSat tables
Since VirSat 4.10 if I select multiple rows in a table (e.g. with Shift+click), only one row is shown as selected, although if I click "Remove", all the objects are removed.

image

ghost:
- Changed the Try runExclusive to also lock the entire workspace
- Added a corresponding testcase

This was needed since the tryRunExclusive was called from some UI Code in update events. This was likely to interfere and deadlock with Workspace interactions, such as the synchronizer. The KatSat DataSet had a lot of dead locks here.

---
Closes #467: Multi-select not shown in VirSat tables


SaMuellerDLR (approved):
Approved!

franzTobiasDLR (approved):

PR #502, opened 2020-02-25
Close all editors - (Task #501)

### LINKED ISSUE #501: Problem with tearDown method in ASWTBotTestCase class
Complete workbench reset causes trouble with SWTBot test case batching

christoph-pietsch-DLR:
Closes all editors instead of resetting the workbench.

Closes #501 
---
Task #501: Problem with tearDown method in ASWTBotTestCase class


SaMuellerDLR:
@christoph-pietsch-DLR Changing the behavior from reset workbenches to a simple close all editors doesnt look like the right step to me.

PR #507, opened 2020-02-27
Adjusting Expression Evaluator in Calculations for Set Functions

ghost:
Set-Functions can such as a summary usually summarize a parameter. How deep they should search in the trees can be restricted. So far the implementation was orienting on the levels of nesting regarding the SEIs and their children. Now it is the the level of occurrence of the property or category in question.

Reused the BeanTraverser for this implementation. Generalized it for the usage of SEIs as well.

Change was needed due to the inconsistency between the summary views and the calculations in CEF.


de.dlr.sc.virsat.model.test/src/de/dlr/sc/virsat/model/dvlm/tree/TreeTraverserTest.java
```diff
@@ -0,0 +1,169 @@
...
+			int result = 1;
+			result = prime * result + Objects.hash(matchingLevel, nestingLevel, sei);
+			return result;
+		}
+
+		@Override
+		public boolean equals(Object obj) {
+			if (this == obj) {
+				return true;
+			} else if (obj == null) {
```
SaMuellerDLR: Would it maybe be sufficient to just check 

`if (obj instanceof Result) {
	Result other = (Result) obj;
	return matchingLevel == other.matchingLevel && nestingLevel == other.nestingLevel && Objects.equals(sei, other.sei);
else {
	return false;
}`

That should cover all the cases.

de.dlr.sc.virsat.model.test/src/de/dlr/sc/virsat/model/dvlm/tree/TreeTraverserTest.java
```diff
@@ -0,0 +1,169 @@
...
+				Result other = (Result) obj;
+				return matchingLevel == other.matchingLevel && nestingLevel == other.nestingLevel && Objects.equals(sei, other.sei);
+			}
+		}
+		
+		Result(StructuralElementInstance sei, int processedLevel, int matchedLevel) {
+			this.sei = sei;
+			this.nestingLevel = processedLevel;
+			this.matchingLevel = matchedLevel;
+		}
```
SaMuellerDLR: Can we move these to the top of the `Result` class so the reader directly knows whats encapsualted by it?

SaMuellerDLR (changes requested):
Looks good. Only have an issue with a minor readability thing in a test case.

de.dlr.sc.virsat.model.test/src/de/dlr/sc/virsat/model/dvlm/tree/TreeTraverserTest.java
```diff
@@ -0,0 +1,169 @@
...
+		
+		traverser.traverse(seiRoot, new IStructuralElementInstanceTreeTraverserMatcher() {
+			@Override
+			public void processMatch(StructuralElementInstance treeNode, StructuralElementInstance matchingParent) {
+				mapToParentResults.put(treeNode, matchingParent);
+			}
+			
+			@Override
+			public boolean isMatching(StructuralElementInstance treeNode) {
+				return treeNode.getName() != null;
```
andriikovalov-dlr: Not immediately clear what it matches. I suggest we either check it explicitly like `return treeNode != seiChild1 && treenode != seiChild2` or check a special name like "MISMATCH", and set it to mismatching elements.

de.dlr.sc.virsat.model/src/de/dlr/sc/virsat/model/dvlm/tree/ITreeTraverserMatcher.java
```diff
@@ -0,0 +1,57 @@
...
+	 */
+	boolean isMatching(TYPE treeNode);
+	
+	/**
+	 * Processes the found match
+	 * @param treeNode a matching tree node
+	 * @param matchingParent the closest parent of the node that also matches the criteria.
+	 *        Can be null if the node does not have a matching parent
+	 */
+	void processMatch(TYPE treeNode, TYPE matchingParent);
```
andriikovalov-dlr: Shall we also add tree level and match level as parameters here?

de.dlr.sc.virsat.model.test/src/de/dlr/sc/virsat/model/dvlm/tree/TreeTraverserTest.java
```diff
@@ -0,0 +1,169 @@
...
+			}
+			
+			@Override
+			public boolean continueTraverseChildren(StructuralElementInstance treeNode, boolean isMatching, int nestingLevel, int matchingLevel) {
+				traverseResults.add(new Result(treeNode, nestingLevel, matchingLevel));
+				
+				// Try to make sure that the seiChild2ChildChild which is theoretically correct, still does not get found
+				boolean isChild2Child = treeNode == seiChild2Child;
+				return !isChild2Child && IStructuralElementInstanceTreeTraverserMatcher.super.continueTraverseChildren(treeNode, isMatching, nestingLevel, matchingLevel);
```
andriikovalov-dlr:
```suggestion
				return treeNode != seiChild2Child;

```

andriikovalov-dlr (changes requested):
A couple of minor comments and suggestions

de.dlr.sc.virsat.model.test/src/de/dlr/sc/virsat/model/dvlm/tree/TreeTraverserTest.java
PhilMFischer: Well, the code present is auto coded by the eclipse ide.  The result is basically the same, but there is a difference in performance. Since performance is not an issue in this test case, and in particular not with this object i don't mind changing it.

de.dlr.sc.virsat.model.test/src/de/dlr/sc/virsat/model/dvlm/tree/TreeTraverserTest.java
PhilMFischer: yes done.

de.dlr.sc.virsat.model/src/de/dlr/sc/virsat/model/dvlm/tree/ITreeTraverserMatcher.java
```diff
@@ -0,0 +1,57 @@
...
+	
+	/**
+	 * This method tells if the children of the current node should be processed or not.
+	 * 
+	 * Default behavior is that all children are processed.
+	 * @param treeNode the node of which the children should be processed or not.
+	 * @param isMatching the matching result which the traverser evaluated before.
+	 * @return true in case the children should be processed. false in case not.
+	 */
+	default boolean continueTraverseChildren(TYPE treeNode, boolean isMatching, int processedLevel, int matchedLevel) {
```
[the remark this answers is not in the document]
PhilMFischer: Good suggestion.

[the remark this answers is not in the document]
PhilMFischer: It is a bit strange checkstyle is not complaining. but probably since the method is that small, it does not expect a javadoc at all.

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/types/structural/tree/BeanStructuralTreeTraverser.java
```diff
@@ -11,55 +11,11 @@
...
-		
-		// Now check if the children should be processed.
-		// if yes recursively loop over all of them.
-		if (matcher.continueTraverseChildren(node, isMatching)) {
-			for (IBeanStructuralElementInstance child : node.getChildren(IBeanStructuralElementInstance.class)) {
-				traverseRecursive(child, nextParent);
-			}
-		}
-	}
+public class BeanStructuralTreeTraverser extends TreeTraverser<IBeanStructuralElementInstance> {
```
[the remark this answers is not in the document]
PhilMFischer: Only for backwards compatibility, because it has been exposed API, and i am not sure about all places where it has been used.

de.dlr.sc.virsat.model/src/de/dlr/sc/virsat/model/dvlm/tree/ITreeTraverserMatcher.java
PhilMFischer: I thought about that one as well, but wanted to keep it to as minimal change as needed. thus i would argues for no, for the moment.

de.dlr.sc.virsat.model.test/src/de/dlr/sc/virsat/model/dvlm/tree/TreeTraverserTest.java
PhilMFischer: The former one, has the default implementation of the interface connected to the test. Thus a change to the default implementation would be captured here. I prefer leaving it as it is, but i added a comment, why the statement has to be that complex.

de.dlr.sc.virsat.model/src/de/dlr/sc/virsat/model/dvlm/tree/ITreeTraverserMatcher.java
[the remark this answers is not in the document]
andriikovalov-dlr: Yes, it's just one line, checkstyle doesn't require a javadoc. But since it's an interface method, it's good to have all parameters explained

SaMuellerDLR (approved):
Approved!

andriikovalov-dlr (approved):

PR #518, opened 2020-03-04
Migration with new dependencies

franzTobiasDLR:
This pull request refactors the activation and migration of conecpt.

Things done:
- Move activation of conept to new dedicated class
- Added tests for activating concepts 
- Fixed some bugs in activating concepts
- Migration now checks if there are new dependencies beofre migrating
- Migration of new features activates these features (didn't happen before :O )

- [x] Done


de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/util/ConceptActivationHelper.java
```diff
@@ -0,0 +1,219 @@
...
+	protected Repository repository;
+	
+	public ConceptActivationHelper(Repository repository) {
+		this.repository = repository;
+	}
+	
+	public ConceptActivationHelper(Concept concept) {
+		if (concept.eContainer() != null && concept.eContainer() instanceof Repository) {
+			this.repository = ((Repository) concept.eContainer());
+		}
```
andriikovalov-dlr: Else throw exception?

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/util/ConceptActivationHelper.java
```diff
@@ -0,0 +1,219 @@
...
+			editingDomain.getCommandStack().execute(cmd);
+		}
+	}
+	
+	/**
+	 * Activate new concept dependencies 
+	 * @param concept the concept to be prepared for migration
+	 * @param editingDomain the editing domain
+	 */
+	public void handleNewDependencies(Concept concept, EditingDomain editingDomain, IProgressMonitor progressMonitor) {
```
andriikovalov-dlr: Should we maybe also handle the situation when a dependency was removed in one of the concept versions?

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/migrator/AMigrator.java
```diff
@@ -204,6 +222,34 @@ protected void removeFromTarget(ReferenceChange diff, boolean rightToLeft) {
 	protected void registerOldToNewIds() {
 	}
 	
+	@Override
+	public Set<String> getNewDependencies(Concept concept, IMigrator previousMigrator) {
+		
+		String conceptId = concept.getFullQualifiedName() + "/";
+		Concept conceptPrevious = loadConceptXmi(conceptId + previousMigrator.getResource());
```
andriikovalov-dlr: According to ther javadoc, `previousMigrator` can be null, which would result in NPE here

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/migrator/AMigrator.java
```diff
@@ -204,6 +222,34 @@ protected void removeFromTarget(ReferenceChange diff, boolean rightToLeft) {
...
+	
+	/**
+	 * Return new dependencies of new concept versions
+	 * @param conceptPrevious the previous concept version. A comparison between conceptPrevious
+	 * and the conceptCurrent should not yield any differences.
+	 * @param conceptCurrent the current concept as it is in the repository
+	 * @param conceptNext the next concept version
+	 * @return A set of new concept names
+	 */
+	public Set<String> getNewDependencies(Concept conceptPrevious, Concept conceptCurrent, Concept conceptNext) {
```
andriikovalov-dlr: Looks like we don't need `conceptPrevious`

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/migrator/AMigrator.java
```diff
@@ -608,6 +656,13 @@ public EObject getEObject(URI uri, boolean loadOnDemand) {
 		String nsURI = ReleaseUtils.getNamespaceURI(conceptResourceUri);
 		Migrator migrator = MigratorRegistry.getInstance().getMigrator(nsURI);
 		
+		if (migrator == null) {
+			DVLMEditPlugin.getPlugin().getLog().log(new Status(Status.ERROR, DVLMEditPlugin.PLUGIN_ID, 
+					"Could not get DVLM migrator for concept resource: " + conceptResourceUri));
+			//Check that all dependent concepts and their plugins are available in platform...  
+			return new ResourceSetImpl();
```
andriikovalov-dlr: Could you please explain this fragment? When can the migrator be null, why are we returning a new resource set and what does it mean to the calling code?

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/util/ConceptActivationHelper.java
franzTobiasDLR: Good idea 👍

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/util/ConceptActivationHelper.java
franzTobiasDLR: How would you handle that? I mean we can add that to the migrator, but we currently do not forsee removing deactivating concepts... So any idea what that would do with the removed dependency?

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/migrator/AMigrator.java
franzTobiasDLR: Well, I had the problem that a test case was failing in the maven build (only in the maven build!) because there the migrator was "null". I was debugging that for several hours until I tried returning a simple resource set in that case... The resource set then informed me that a required concepts of that cocept could not be found, because its plugin was not found... So I simply had to add the dependent concept's plugin as required plugin in the test fragement...
If you don't like that we can also remove this fragement... But I was that pissed of that the migrator was just "null" without any explonation that I added this pease of code. I had that pronblem some time ago and I think there that would have helped as well

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/migrator/AMigrator.java
franzTobiasDLR: Oh true, and actually the same holds for the migration itself

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/migrator/AMigrator.java
franzTobiasDLR: Yes, I just wanted to keep the interface analog to the migration... But I guess that only overcomplicates thinks

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/util/ConceptActivationHelper.java
andriikovalov-dlr: Assume v1 has no dependency, v2 has a dependency, v3 has no dependency again. If I want to migrate from v1 to v3, I don't need to activate this dependency. But I guess it's not going to happen often, so we leave it

de.dlr.sc.virsat.model.edit.test/src/de/dlr/sc/virsat/model/concept/migrator/AMigratorTest.java
```diff
@@ -136,6 +139,8 @@ public void testMigrateAddStructuralElement() {
 		Concept conceptPrevious = ConceptsFactory.eINSTANCE.createConcept();
 		Concept conceptCurrent = ConceptsFactory.eINSTANCE.createConcept();
 		Concept conceptNext = ConceptsFactory.eINSTANCE.createConcept();
+		Repository repository = DVLMFactory.eINSTANCE.createRepository();
+		repository.getActiveConcepts().add(conceptCurrent);
```
PhilMFischer: Maybe a little hint why the concept has to be placed into the repository?

franzTobiasDLR:
> 
> 
> Still a question from my side:
> How does the selection dialog now react if i select a concept which depends on another one? Are required concepts directly selected in the dialog, or do they just become visible after you accepted to install them to the repo?

The behavior of the concept selection dialog didn't change... Required concepts are automatically selected - but not added automatically.

PR #522, opened 2020-03-05
Add .empty files to VirSatProjectCommons folders - (Task #521)

### LINKED ISSUE #521: Adjust VirSatProjectCommons for git
Every created folder has an empty .empty file for git to detect the folder.

This change is needed for the new version control system architecture in #519.

E.g. git will only put a directory under version control if a file exists inside.

JAmmermann-DLR:
For git to detect empty folders.

Closes #521

---
Task #521: Adjust VirSatProjectCommons for git


JAmmermann-DLR:
SWT bot crashes in CutCopyDeleteUndoTest in undo(...):
- because in WorkspaceUndoUtil  recreate(...) resourcesToRecreate is null
- because in DeleteResourcesOperation doUndo(...) uiInfo is null

Not sure how this is related to the added . empty file, which definitively is the cause of this problem somehow.
Maybe at some point not the sei file get's read in but the .empty file instead.

PhilMFischer:
The problem was, that the .empty files were create with the java file api but not he eclipse resource api. Thus eclipse is not correctly aware of these files creating out of sync events which create trouble on delete and undo operations since eclipse cannot tell what to do with these changes on the file system.

PR #563, opened 2020-04-05
Improve editing domain's handling of Non-DVLM Models 

### LINKED ISSUE #562: Inconsistency in transactional builders with Non-DVLM resources
Our transactional builders (inheritance + equation) consider changes only if they're in DVLM resources... (Which makes sense). After building they use the editing domain's saveAll method and thus save all resources in the editing domain's resource set. That couses problem:

Non-DVLM Resources that have been changed on the filesystem are not loaded but saved. This inconsitency makes the builders overwriting the non-dvlm resource all the time. Resources cannot even be deleted because the build always saves them and thus brings them back.

Possible solutions could be to always load all resources that are in the resource set or to only save the resources that are loaded by the builders.

franzTobiasDLR:
Closes #562 

Currently the transactional builders save all resources in the resource set after building. The equation and inheritance builder (and also the ResourceChangeListner) however ignore non-dvlm changes and thus then overwrite external changes. As it makes sense that equation + inheritance builder only work with DVLM resources and it also make sense that the saveAll() on the TransationalEditingDomain saves all resources, I added an option to only save DVLM resources which can be used from builders. 

- Add a saveAll option to only save DVLM resources in the editing domain
- Add option in transactional builders to uses this new saveAll option
if the builder only works with DVLM resources
- Update inheritance and equation builder to only save DVLM resources
-Update resource change listner to also handle changes of non-dvlm files


franzTobiasDLR:
> 
> 
> I agree to these changes.
> 
> The VirSat Builder should only change DVLM resources and thus save them. This maybe a different case for specialized builders, which have more knowledge about other models referencing into DVLM, but his is not the case here.
> 
> Still it looks to me as if the TransactionalEditingDomain should reload non DVLM resources in an associated resourceSet as you already described in #562. Otherwise a dvlm referencing an external ecore would not reload an external change and overwrite it next time the resourceSet is saved.

True! And as we implemented a TransactionalEditingDomain and not a DVLMEditingDomain it also makes sense that it also handles non-DVLM models. I updated it so that it also reloads other model types, if they are in the editting domain's resource set.

de.dlr.sc.virsat.model.extension.tests.test/src/de/dlr/sc/virsat/project/editingDomain/VirSatTransactionalEditingDomainNonDVLMTest.java
```diff
@@ -0,0 +1,127 @@
...
+		// Create external model content
+		containerVirsatContext = TestsFactory.eINSTANCE.createContainer();
+		executeAsCommand(() -> externalModelResourceVirSatDomain.getContents().add(containerVirsatContext));
+		externalModelResourceVirSatDomain.save(Collections.emptyMap());
+	}
+	
+	@Test
+	public void testSaveAll() throws IOException {
+		// Do some external changes
+		TransactionalEditingDomain externalEditingDomain = TransactionalEditingDomain.Factory.INSTANCE.createEditingDomain();
```
andriikovalov-dlr, line 70: Let's move repetitive code into a method like `doExternalChange()`

Issue #548, opened 2020-03-31
Introduce capabilities to override current user for command executions


PhilMFischer:
The server communicates with the backend using a functional account. Nevertheless users have to authenticate to the server with their user password credentials. When a user is now writing to the data model, the authenticated user should be forwarded to the RoleManagementChecks. Therefore the RoleManagement - Architecture needs to be refactored since it currently refers to the current system user.

PR #571, opened 2020-04-16
PR #1 for User Context Implementation - DVLM Adjustments

ghost:
This is PR no1 for a series of PR to adjust the user management and to change it from a single user one for desktop applications to a multi user one as needed for the rest server application.

The general idea is to introduce a user context which is using the system user within a desktop application. Within the server the http authentication is handing over the user to the context and executing it atomically in the editing domain. One goal is to still provide the intrinsic rights management on the model, even without an editing domain.

This PR focuses on the changes on the data model. The following points are done:
- Added new _IUserContext_ Interface. This interface is handed over to the _RightsHelper_. The _RighstHelper_ has all the code to judge if a user can write to a part of the data model or not. The _UserRegistry_ which reads the user from the system is now implementing the _IUserContext_ Interface.
- _RoleManagementCheckCommand_ is prepared to either detect the _IUserContext_ from the domain or use the system one. All related _model.edit_ code including the templates have been adjusted and regenerated.

---
Task #548: Introduce capabilities to override current user for command
executions


SaMuellerDLR (approved):
Approved!

franzTobiasDLR (approved):

PR #578, opened 2020-04-17
Improve Test Case for Git Backend

### LINKED ISSUE #577: Instability at Git VersionControlBackend Tests
Executing the Backend Tests for Git sometimes fail in the IDE.

Apparently the problem is due to process locking when deleting projects.

ghost:
- Added an abstract method for an interim check to proof the remote repo
is correct.
- Added test method to allow for retries on an assert statement
- Improved project tear down, to infinetly try to remove projects, to
avoid problems with inter-process locking.

---
Closes #577: Instability at Git VersionControlBackend Tests


ghost (approved):

SaMuellerDLR (approved):
Approved!

PR #583, opened 2020-04-21
Fixing shaky test cases - Adding Test Cases for Excel Exporter Helper

### LINKED ISSUE #582: Fix shakey code coverage in ExcelExportHelper
The Code in the Excel Export Helper has some shakey behaviour due to the use of _LocalDateTime.now()_. The date time should maybe handed over from external. The test case should test both cases and should not have a direct system time dependency.

```
LocalDateTime ldt = LocalDateTime.now();
row.getCell(datacell).setCellValue(ldt.getDayOfMonth() + "/" + ldt.getMonthValue() + "/" + ldt.getYear());

row = headerSheet.getRow(AExcelIo.HEADER_ROW_TIME);
if (ldt.getMinute() < ADDZEROIFLESS) {
	row.getCell(datacell).setCellValue(ldt.getHour() + ":0" + ldt.getMinute());
} else {
	row.getCell(datacell).setCellValue(ldt.getHour() + ":" + ldt.getMinute());
}
```

image

ghost:
- Minor code clean up on helper such as renaming methods.
- Added specific test cases for the helper class.

@SaMuellerDLR I had the impression as if the exporters are not used at all in the Virtual Satellite code. I think you and @ALEhmerDLR were the last ones working in this area. What is your opinion?

---
Closes #582: Fix shakey code coverage in ExcelExportHelper

### DIFF
--- a/de.dlr.sc.virsat.excel.test/src/de/dlr/sc/virsat/excel/exporter/ExcelExportHelperTest.java
+++ b/de.dlr.sc.virsat.excel.test/src/de/dlr/sc/virsat/excel/exporter/ExcelExportHelperTest.java
@@ -1,0 +1,110 @@
+/*******************************************************************************
+ * Copyright (c) 2008-2019 German Aerospace Center (DLR), Simulation and Software Technology, Germany.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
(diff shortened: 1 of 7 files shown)


SaMuellerDLR (approved):
Changes look good to me.

@PhilMFischer Do mean they are not used in the code because they are not called directly by any method? If so: The exporters / importers implementations are registered using extension points in the respective plugin.xml.

PhilMFischer:
> Changes look good to me.
> @PhilMFischer Do mean they are not used in the code because they are not called directly by any method? If so: The exporters / importers implementations are registered using extension points in the respective plugin.xml.

Right, I see. Makes sense.

franzTobiasDLR (approved):

PR #590, opened 2020-04-29
Performance improvements by reducing number of global ResourceSet actions

### LINKED ISSUE #589: Improve performance by reducing global resource set operations
We're having quite a lot of DVLM operations that work on ResourceSet level, thus touching all resource in the resource set. In combination with all our ResourceSet-level resolveAlls this is quite bad for the performance as it triggers an incredibale amout of events in the workspace.

franzTobiasDLR:
We're using global ResourceSet actions guite a lot in Virtual Satellite. Such global actions are performed on all resources in the ResourceSet and even trigger loading resources of referenced objects. While this is intended for DVLM files it is also performed on non-DVLM files and can reduce the performance quite a lot. 

Removing dangeling references, collecting types on ResourceSet-level and the validation, equation + inthertiance builders currently all resolve all resouces in the ResourceSet. With resources for diagrams and DSLs in the ResourceSet, this triggers long chains of resource events, because also references into diagram infrastucutre and so on are resolved. 

This PR tries to reduce overhead by using actions on DVLM files only where non-DVLM files are not needed. Such as e.g. getting all SEIs in a project.

Closes #589


de.dlr.sc.virsat.project/src/de/dlr/sc/virsat/project/structure/VirSatProjectResourceChangeListener.java
```diff
@@ -122,10 +122,12 @@ public boolean visit(IResourceDelta delta) {
 			// workspace job as we will need the editing domain to handle resource changes.
 			// This way the current job calling us can safely finish and the new job
 			// will handle the actual changes.
-			Job job = new WorkspaceSynchronizerJob(addedDvlmResources, removedDvlmResources, changedDvlmResources, counter++);
-			String changeList = printLists(addedDvlmResources, removedDvlmResources, changedDvlmResources);
-			Activator.getDefault().getLog().log(new Status(Status.INFO, Activator.getPluginId(), "VirSatProjectResourceChangeListener: Scheduling " + job.getName() + ". \n" + changeList));
-			job.schedule();
+			if (!addedDvlmResources.isEmpty() || !removedDvlmResources.isEmpty() || !changedDvlmResources.isEmpty()) {
```
PhilMFischer, line 125: This is also a dificult one. The whole story to the workspace synchtronizer is, to detect if there are changes to a reosurce of the reosurceset which have been introduced somewhere on the workspace. In that case, the reosurce is out of sync with the loaded reosurce. EMF usually handles this in the generated editors. In VirSat we decided to handle this centrally, to also manage intended changes. E.g. when a resource is explicitly saved, which creates a workspace resource change event as well. These intended changes should obviously not lead to a resource reload. So at least we need to have a convention, agreement how to deal with non dvlm resources. In this proposed implementation, specific editors have to deal with reloading the resources.

de.dlr.sc.virsat.project/src/de/dlr/sc/virsat/project/resources/VirSatResourceSet.java
```diff
@@ -1151,7 +1175,9 @@ public Resource createResource(URI uri) {
 	 */
 	public boolean updateDiagnostic(Resource resource) {
 		boolean changes = false;
-		if (resource != null) {
+		// Only run diagnostic on DVLM files as other resource might result in a huge chain of
+		// resource that have to be resolved (e.g. diagrams also have references to the diagram infrastructure...)
+		if (resource != null && isDvlmResource(resource)) {
```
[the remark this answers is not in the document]
franzTobiasDLR: Yes, to be onest I wouldn't expect an editing domain to automatically check for resource problems. If you have broken files on your computer's file system, then you still see them only when trying to open the file... And problems in resources usally happen if something goes wrong in the editor... 
For me the most problematic point here is, however, that this operation resolves all links... also to unloaded files. Thus, in each change on the file system (e.g. save) there are lots of IO operations going on. I saw cases where suddenly over 300 resources were loaded... Then nothing worked anymore.
I like that we're handling resource problems for our DVLM files, but IMHO that only works with a clear defined scope. An editing domain, on the other hand, should not be bound to a limited file scope.
If you prefer on also checking non-DVLM resources for problems I would rather do that on loaded resources only... Especially as resources such as LazyLinkingResources are desgined to be lazy-linking... Just imagine a browser would try to resolve all references in the internet.... :D If you have a differnt option let's maybe discuss that in aphone call ;)

de.dlr.sc.virsat.project/src/de/dlr/sc/virsat/project/resources/VirSatResourceSet.java
```diff
@@ -1297,4 +1323,15 @@ public void loadAllResources() {
 		getRepositoryResource();
 		EcoreUtil.resolveAll(this);
 	}
+	
+	/**
+	 * Resolve all DVLM resources in this resource set
+	 */
+	public void resolveAllDvlmResources() {
```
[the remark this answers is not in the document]
franzTobiasDLR: I just considered it to be related to the loadAllResources() in the resources set... But I can of course also move it...

de.dlr.sc.virsat.project/src/de/dlr/sc/virsat/project/structure/VirSatProjectResourceChangeListener.java
franzTobiasDLR: I have no problem at all iwth the WorkspaceSynchronizer. I actually like it :) I just want to make sure that is only triggered if there is actually a relvant change on the project. With the current implementation the WorkspaceSychonizer run into an enless loop when I linked resources into a VirSat project...

de.dlr.sc.virsat.project.test/src/de/dlr/sc/virsat/project/resources/VirSatResourceSetTest.java
```diff
@@ -588,6 +591,131 @@ public void testLoadAllResources() {
...
+		VirSatResourceSet resSet = VirSatResourceSet.createUnmanagedResourceSet(testProject);
+		resSet.initializeModelsAndResourceSet();
+		
+		Repository repo = resSet.getRepository();
+		
+		StructuralElement se = StructuralFactory.eINSTANCE.createStructuralElement();
+		se.setIsApplicableForAll(true);
+		
+		//CHECKSTYLE:OFF
+		StructuralElementInstance sei1 = StructuralFactory.eINSTANCE.createStructuralElementInstance();
```
SaMuellerDLR, line 605: Can probably do this a bit simpler with a `StructuralInstantiator`

de.dlr.sc.virsat.project.test/src/de/dlr/sc/virsat/project/resources/VirSatResourceSetTest.java
```diff
@@ -588,6 +591,131 @@ public void testLoadAllResources() {
...
+		resSei2.unload();
+		resSei2_1.unload();
+		resSei2_1_1.unload();
+		
+		URI uri = URI.createPlatformResourceURI("/testProject_VirSatResourceSetTest/test.ecore", true);
+		Resource newResource = new XMIResourceImpl(uri);
+		newResource.save(Collections.EMPTY_MAP);
+		Resource nonDvlmResourceReload = resSet.getResource(uri, true);
+		
+		assertTrue("Should contain DVLM resource", resSet.getDvlmResources().contains(resSei1));
```
SaMuellerDLR, line 711: We can probably simplify this with `assertThat(message, resSet.getDvlmResources(), hasItems(...)`

PR #613, opened 2020-05-28
Updated External Lib Jersey build process  and fixed moxy dependencies

### LINKED ISSUE #611: Implement maven supported Dependency Management for external libraries
Our external libraries include Jars which are manually copied into the plugins. The class path exposes them in the plugin. This approach is needed for the maven/tycho builds.

In the case of jersey for example, the copied jars are from a downloadable bundle, which are missing some dependencies to the MOXy refernce implementation used for json transformation. This is a problem, since the dependencies and their transistive dependencies need to be copied by hand.

The idea is to use pure maven to get the dependencies for the plugin and copy them into the external lib folder. Additionaly the attached licenses could be taken from maven as well to create an overview of used licenses.

ghost:
- Updated Jersey Dependencies to use Moxy for json transformations

- Had to work with maven dependencies to get correct libraries.
- Updated library build process to allow for lib download from maven repositories.
- Third party libraries can be added as maven dependencies
- The profile "downlaodLibs" copies the libs into the extrenal folder and generates a Thrid Party License file as well
- Created and inetr parent pom which can be used for these maven based lib plugins.
- also updated the about.html to fit to the adjusted structure and process.

- [x] All Done.

Closes #611


SaMuellerDLR:
@PhilMFischer server test cases seem to be failing on travis

franzTobiasDLR (approved):

ghost (approved):

SaMuellerDLR (approved):
Approved!

SaMuellerDLR:
@PhilMFischer whats going on with the task lisk check? :D

PR #619, opened 2020-06-03
Sortable columns in QUDV Editor

### LINKED ISSUE #30: Make QUDV Editor tables sortable
The QUDV Editor tables are not sortable by now.

Every Column should be sortable. Standard Sorting should be by Ascending Name.

QUDV_Editor_Tables

SaMuellerDLR:
Implemented new LabelColumnSorter that provides generic functionality to sort a table according to 
* a selected column by the user,
* a sort direction (ascending, descending),
* and the label of the objects.

Small refactoring in generic snippets to exract creation process of a table viewer to make it easier to override just that logic instead of the entire setup process of the table viewer

Have some funny picture:

sorting

By default the tables are sorted ascending by Unit Name and Quantitiy Kind Name.

Apparantly you cant customize the placement of the sorting indicator (the arrow thing), as that seems to be somehow determined by the operating system.

Closes #30

### DIFF
--- a/de.dlr.sc.virsat.project.ui/src/de/dlr/sc/virsat/project/ui/navigator/commonSorter/LabelColumnSorter.java
+++ b/de.dlr.sc.virsat.project.ui/src/de/dlr/sc/virsat/project/ui/navigator/commonSorter/LabelColumnSorter.java
@@ -1,0 +1,71 @@
+/*******************************************************************************
+ * Copyright (c) 2020 German Aerospace Center (DLR), Simulation and Software Technology, Germany.
+ *
+ * This program and the accompanying materials are made available under the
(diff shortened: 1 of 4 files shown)


SaMuellerDLR:
@PhilMFischer I checked the implementations of ViewerComparator and there seems to be no standard implementation for it in the swt/jface plugins. I found a very similar looking implementation in egui.ui though :D 

Regarding all other tables such as seis/category assignments: There we currently use the order as they are in the navigator. We would need to support switching between the ascending/descending/default ordering. Which should not be too difficult. But I believe it would require a bit more refactoring. Similarly, I think for storing and retrieving the sorter state, we would need some more refactoring, since the existing code there is heavily geared towards just saving/restoring the expansion state.

PR #646, opened 2020-06-17
Fixed method calls in Editing domain - (Task #645)

### LINKED ISSUE #645: Fix writeExclusive method in Editing Domain to lock workspace
The method writeExclusive is the counterpart to runExclusive.

For VirtualSatelltie the runExclusive is locking the workspace first before going into the transaction.
This part is missing for the writeExclsuive part and can cause deadlocks.

ghost:
Fixed an issue with the Transactional Editing Domain, where the writeExclusive could be called outside of the workspace. This was leading to deadlocks.

Anyway, not sure if we understand the transaction mechanism in full depth since the writeExclusive is implemented by us. We should discuss if we completly remove the method and stick with the run exclsuive from the EMF transactional framework. It creates a read lock. Not sure if the read lock will also avoid writing with otehr transactions, or if transaction are never in parralel anyway.

---
Closes #645: Fix writeExclusive method in Editing Domain to lock workspace

### DIFF
--- a/de.dlr.sc.virsat.model.calculation.ui/src/de/dlr/sc/virsat/model/calculation/ui/editor/EquationSectionUriEditorInput.java
+++ b/de.dlr.sc.virsat.model.calculation.ui/src/de/dlr/sc/virsat/model/calculation/ui/editor/EquationSectionUriEditorInput.java
@@ -338,8 +338,6 @@ public class EquationSectionUriEditorInput extends FileEditorInput implements IP
 	 */
 	public void saveResource() {
 		if (editingDomain.isDirty(virSatResource)) {
-			// Doing the save in a workspace operation is intended, to not let
-			// the automatic builders jump on these files in between
 			editingDomain.saveAll();
 		}
 	}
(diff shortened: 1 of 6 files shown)


PhilMFischer:
Removing the writeExclusive is not possible. We checked that there are some actions such as saving which need to run in an workspaceoperation but still need to execute a command, thus asking for a writabel transaction and not just a reading one.

Anyway we additionally clean up the code by:
- Removing WorkspaceModifyOperations which were incorrectly applied and not necessary anymore, since the workspace lock is now an intrinsic part of the TransactionalEditingDomain.
- Cleaned up the use of workspace and transaction locking in the TransactionalEditingDomain

SaMuellerDLR (approved):
Approved!

de.dlr.sc.virsat.model.calculation.ui/src/de/dlr/sc/virsat/model/calculation/ui/editor/EquationSectionUriEditorInput.java
```diff
@@ -345,21 +340,7 @@ public void saveResource() {
 		if (editingDomain.isDirty(virSatResource)) {
 			// Doing the save in a workspace operation is intended, to not let
```
franzTobiasDLR, line 341: Guess these comments are outdated then? Is that no concern anymore? "to not let the automatic builders jump on these files in between"

SaMuellerDLR: The saveAll of the editing domain was changed to lock the workspace before doing anything else (which is why the workspace modify operation is no longer needed), so I do think the comment can also go, yes.
I will quickly remove it.

franzTobiasDLR (approved):

PR #672, opened 2020-06-26
Exchanged write access checks in snippet for equation section

### LINKED ISSUE #374: Add ability to see equations of not owned elements
Currently the only way to see the equations is via "Edit equations", and it is only available for the owner. We should add the possibility for everyone to view the equations, but not edit.

SaMuellerDLR:
- The Equation section edit button is no longer checked for rights
management
- The perform update button is now checked for rights management

Closes #374

---
Task #374: Add ability to see equations of not owned elements


PhilMFischer:
@SaMuellerDLR With this change it is possible to open the editor now. Is the editor avoiding save in case the user does not have the rights?

SaMuellerDLR:
@PhilMFischer When a user without rights opens the equation section, the editor pops up and he can write it in but pressing save does _not_ carry over the changes into the data model.

PR #475, opened 2020-02-17
Fixed TableNavigation and Selection 

ghost:
- The Table Navigation corrupted the selection highlighting. This has
been adjusted.
- Additionally the behavior for editing a cell has been changed, so that
the cell editors are not directly activated, but on double-click or two
sequential single-clicks into the same cell.
- Adjusted CellEditors using DialogCellEditos. Corrected "tab"
redirection into the cell.

---
Task #467: Multi-select not shown in VirSat tables

Closes #467 - Also cherry picked into integration for Release 4.11.0

PR #674, opened 2020-06-29
Synchronizing editor resource reload in workspace operation

### LINKED ISSUE #671: Synchronize Editor Resource Reload with Workspace
A reload of the underlying resource in the Generic Editor is executed by an asynchronous execution of the display thread.

This emans execution time is unpreditable and may happen during a load or or unload of resources. Thus the exceution should be interlocked with the workspace.

ghost:
- Synchronized resource releoad in GenericEditor with workpsace using the runExclusive methods from the editing domain. This avoids editor updates in case, that a builder or an external change is altering the resource set while loading and unloading. This is also a primary suspect for editors unexpectedly closing and dissapearing in SWTBot Tests.
- Also added a SWTBot test to check the editor gets closed as e expected if an external change (e.g. removing the resource from the file system) happens.

DUe to the complexity of this change and reoccuring problems in swtbot test cases the following parts of VirtualSatellite are now tested as well by new testc ases:
- The abstract ResourceChangeEventListener that checks if a reosurce has been added, removed or changed, taking the recentlysaved state into account.
- The notficications triggered by the actual implementation of the ResourceChangeListener in the TransactionalEditingDomain
- Added general improvements for waiting on notifications both in the ED and the SWTBot test cases

Closes #475


SaMuellerDLR (approved):
Approved!

franzTobiasDLR (approved):
Nice 👍  Looks like you added quite some usefull tests to the ResourceChangeListener

Issue #679, opened 2020-07-03
Generate CategoryBean files


JAmmermann-DLR:
Generate files from GenerateCategoryBeans with the changes from:
- https://github.com/virtualsatellite/VirtualSatellite4-Core/issues/486
- https://github.com/virtualsatellite/VirtualSatellite4-Core/issues/663
- https://github.com/virtualsatellite/VirtualSatellite4-Core/issues/676

And resolve errors based on the API changes

PR #686, opened 2020-07-06
Generate code and fix errors in test cases - (Task #679)

JAmmermann-DLR:
Regenerated model code for all core concepts.

---
Task #679: Generate CategoryBean files


JAmmermann-DLR:
We discovered a future problem when we upgrade to the new Eclipse IDE:
There will be warnings in the generated files for imports in the same package. This could be resolved with changing the ImportManager to check if a class is in the same package before adding it to the imported classes.
grafik

PR #692, opened 2020-07-08
Feature/task 684 enable filtering of available objects in reference selection dialog

### LINKED ISSUE #684: Enable filtering of available objects in ReferenceSelectionDialog
The ReferenceSelectionDialog is the dialog that is opened when you click on the "..." button of an objects property to select the object that it should point to. You can already change the root SEI that the Dialogs ContentProvider gets all of the objects available for selection from.

This issue is about further restricting the objects that a user can select by filtering them with some custom criteria.

watollah-dlr:
New pull request for the same issue. Based on your feedback for the first one.

Closes #684


SaMuellerDLR:
@watollah-dlr By the way please try to commit with your github credentials, so github can properly connect the commits to your account here as well.

PR #696, opened 2020-07-14
Expand navigator to target when drag & drop - (Task #31)

### LINKED ISSUE #31: Expand navigator on drag and drop from concept palette
When dropping a new object from the concept palette, the target object in the navigator does not get expanded.

Improved bahaviour suggests expanding the navigator to the element where the new item got dropped to. Otherwise users are not aware of if there object got added or not.

Anna-Lena-DLR:
When dropping from the concept palette to navigator, the navigator expands to
target level.

closes #31 
---
Task #31: Expand navigator on drag and drop from concept palette


Anna-Lena-DLR:
@PhilMFischer, @SaMuellerDLR  should the newly added item also be selected in the navigator? So in case there are for example more than one product trees, the user know which one he just added?

grafik

SaMuellerDLR:
@Anna-Lena-DLR I think its fine like this. Is this still WIP? Changes look good to me.

Anna-Lena-DLR:
@SaMuellerDLR no then it's fine. I just put the WIP because of my question.

andriikovalov-dlr (approved):

Issue #705, opened 2020-07-20
Improve handling of static lists in AArrayInstanceList


JAmmermann-DLR:
Currently, many methods of `AArrayInstanceList` just throw an `UnsupportedOperationException` when the list is static. Instead a pointer should be added and the `AArrayInstanceList`should allow operations, e.g. adding of an element, as long as the number of elements does not exceed the size of the list/the pointer points to an existing element in the list.

These changes are needed for more sufficient marshalling of lists in https://github.com/virtualsatellite/VirtualSatellite4-Core/issues/607.

PR #706, opened 2020-07-20
Updated handling of set and add in TypeSafeArrayInstanceList - (Task #705)

JAmmermann-DLR:
Modified `TypeSafeArrayInstanceList` instead of `AArrayInstanceList` because the `AArrayInstanceList` works on any object, but properties of the bean are needed. 
Modifications:
- Now set is also allowed for static lists, but only if the new object has the same uuid as the old one.
- Now add on a static lists calls a set if an object with the same uuid exists in the list already.
- Updated the test case.

---
Task #705: Improve handling of static lists in AArrayInstanceList

### DIFF
--- a/de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/list/TypeSafeArrayInstanceList.java
+++ b/de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/list/TypeSafeArrayInstanceList.java
@@ -14,6 +14,7 @@ import java.util.Collection;
 import java.util.List;
 
 import de.dlr.sc.virsat.model.concept.types.property.IBeanProperty;
+import de.dlr.sc.virsat.model.dvlm.categories.propertydefinitions.StaticArrayModifier;
 import de.dlr.sc.virsat.model.dvlm.categories.propertyinstances.APropertyInstance;
 import de.dlr.sc.virsat.model.dvlm.categories.propertyinstances.ArrayInstance;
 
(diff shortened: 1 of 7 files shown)


JAmmermann-DLR:
Things I am not sure of yet:
- I made the set method available for all static lists, because for me a set makes perfectly sense to set elements also on static lists.
- I updated the add method, so that it uses an internal pointer. Did I implement it the way you thought of it? I am not sure if this actually solves the problem we have with the list. We have to make sure initially the right pointer is set (e.g. 0 if we add all the elements when unmarshalling), so I think we probably still need an adapter to ensure this, right?
- I thougt of changing other methods, but actually I don't want to further implement behavior contradictory to the `List` interface:    
  - I think we could also change `addAll(collection)` using the pointer for internal consistency. 
  - For `remove(element)` we don't need a pointer, maybe we still want to implement it.
  - For `addAll(index)`, `remove(index)` and `add(index)` the `List` inteface would expect a shift of elements and then the operation, but the shift makes no sense to me on static lists. Or we have to ensure we actually can shift all elements regarding the size of the static list.

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/list/TypeSafeArrayInstanceList.java
```diff
@@ -191,14 +192,53 @@ public BEAN_TYPE get(int index) {
...
+	/**
+	 * Custom handling for static lists:
+	 * Checks if an element with the same UUID exists in the list
+	 * If that is the case the set method is called,
+	 * else an UnsupportedOperationException is thrown
+	 */
+	@Override
+	public boolean add(BEAN_TYPE e) {
+		if (arrayModifier instanceof StaticArrayModifier) {
+			int index = -1;
```
SaMuellerDLR, line 223: Lets move this block into a small helper method like getIndexByUUID(...).

de.dlr.sc.virsat.model.extension.tests.test/src/de/dlr/sc/virsat/model/extension/tests/model/TestCategoryIntrinsicArrayStaticTest.java
```diff
@@ -80,20 +80,30 @@ public void testAddAllIntCollectionOfQextendsBeansType() {
...
+			arrayStatic.add(property1);
+		});
+		
+		property1.getTypeInstance().setUuid(arrayStatic.get(0).getTypeInstance().getUuid());
+		assertNotEquals("Elements are not the same", property1.getTypeInstance(), arrayStatic.get(0).getTypeInstance());
+		arrayStatic.add(property1);
+		assertEquals("Added the element", property1.getTypeInstance(), arrayStatic.get(0).getTypeInstance());
+		
+		arrayStatic.add(property1);
+		assertEquals("Adding the same element again is be idempotend for static lists", property1.getTypeInstance(), arrayStatic.get(0).getTypeInstance());
```
SaMuellerDLR, line 99: "is be" looks redundant

de.dlr.sc.virsat.model.extension.tests.test/src/de/dlr/sc/virsat/model/extension/tests/model/TestCategoryIntrinsicArrayStaticTest.java
```diff
@@ -1,170 +1,186 @@
...
-	}
-
-	@Test(expected = UnsupportedOperationException.class)
-	public void testSetIntBeanType() {
-		BeanPropertyString property3 = createNewStringProperty();
-
-		arrayStatic.set(1, property3);
-	}
-}
+/*******************************************************************************
```
andriikovalov-dlr, line 1: The whole file is marked as changed - probably wrong line endings? Let's revert it and only have meaningful changes to have a cleaner history

JAmmermann-DLR: I think I deleted some trailing tabs -.- Now you can see the changes.

PR #711, opened 2020-07-23
Git and SVN backends now properly resolve merge conflicts and dialogs show backend errors

### LINKED ISSUE #708: Problems with GitBackend deleting files and quietly denying push
Detected two Problems regarding the git backend on current 4.12.0 integration.

Problem one is about calling commit. In case a new commit is on the remote repository, the "commit" from virtual satellite executes the commit to the local repo, but cannot push. This is expected git behavior but it is not reported to the user.

Problem two is about updateing your repository. In case you are e.g. adding a new SEI and another one commited a change to the remote repository. The git backend is currently applying the MergeSTrategy.THEIRS. Even indended for conflicting files, this approach is a bit more radical and completly replaces the tree with THEIRS. This means as soon as a merge needs to be done all local changes are gone.

SaMuellerDLR:
- The simple THEIRS strategy for the git backend is replaced by the default strategy RECURSIVE + all conflicting files are replaced with the remote version. As a result, instead of discarding all local changes, only conflicting changes are discarded in a merge.
- SVN now calls resolve to replaced conflicted files with the remote version
- Added new test case to check that a.) conflict resolution is applied correctly in both backends (replace with remote in case of conflict) and b.) only conflicting files are affected by this.
- Added EGit dialogs that show the push and pull results for error feedback
- Adapted SWTBot test cases to close the new EGit dialogs

Closes #708

### DIFF
--- a/de.dlr.sc.virsat.team/src/de/dlr/sc/virsat/team/git/VirSatGitVersionControlBackend.java
+++ b/de.dlr.sc.virsat.team/src/de/dlr/sc/virsat/team/git/VirSatGitVersionControlBackend.java
@@ -11,6 +11,7 @@ package de.dlr.sc.virsat.team.git;
 
 import java.io.File;
 import java.util.Collection;
+import java.util.Set;
 
 import org.eclipse.core.internal.resources.Resource;
 import org.eclipse.core.resources.IProject;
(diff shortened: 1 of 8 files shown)


de.dlr.sc.virsat.team/src/de/dlr/sc/virsat/team/git/VirSatGitVersionControlBackend.java
```diff
@@ -93,41 +121,87 @@ protected void doCommit(Repository gitRepository, String message, SubMonitor mon
...
+			checkoutCommand.setProgressMonitor(gitCheckoutMonitor);
+			
+			for (String conflictingFile : conflictingFiles) {
+				checkoutCommand.addPath(conflictingFile);
+			}
+			checkoutCommand.setStage(Stage.THEIRS);
+			checkoutCommand.call();
+			
+			// Finally close the still open merge commit
+			doCommit(gitRepository, BACKEND_REPOSITORY_MERGE_COMMIT_MESSAGE, progressMonitor.split(1));
```
franzTobiasDLR, line 190: Do we maybe want to notify the user about the replacement? Like with a simple dialog? I'm afraid people will not realize that the change was reverted...

franzTobiasDLR (approved):
Nice 👍

de.dlr.sc.virsat.team/src/de/dlr/sc/virsat/team/git/VirSatGitVersionControlBackend.java
SaMuellerDLR: @franzTobiasDLR Its not much but in the EGit dialog you at least see the status notification that there were conflicts.

andriikovalov-dlr (approved):

PR #715, opened 2020-07-29
Fixing Incremental Builder on external runs - (Task #714)

### LINKED ISSUE #714: Changing the Resource on a modeled document does not set the editor in a dirty mode
When adding a new file to a document a file is usually copied into the current workspace. This causes the  builders to run an incremental build on that particular file. Even though no sei got touched, it gets saved by the builder.

Actually the builder should not react to files being non dvlm.

ghost:
- Added a SaveTrigger to incremental builder to individually set if the
whole ResourceSet should be saved.
- Adjusted InheritacneBuilder to only save on changes on DVLM reosurces
- Fixed CalculationBuilder to only save on changes on DVLM Reosurces.
- Added a new test case to check the save trigger.

---
Closes #714: Changing the Resource on a modeled document does not set the
editor in a dirty mode


SaMuellerDLR (approved):
Approved!

franzTobiasDLR (approved):

Issue #726, opened 2020-08-25
Missing Editing domain in REST Server


JAmmermann-DLR:
When the server ist started via the run configuration, existing projects with already checked out repository don't have an editing domain set. This causes internal server errors (500) when accessing the projects model via the api.

Maybe related to https://github.com/virtualsatellite/VirtualSatellite4-Core/issues/556.

PR #727, opened 2020-08-26
Added missing ed for existing projects to ServerRepository- (Task #726)

JAmmermann-DLR:
Now retrieve the ed if a project is already checked out in the ServerRepository

---
Closes Task #726: Missing Editing domain in REST Server


SaMuellerDLR (approved):
That looks simple enough. Approved!

andriikovalov-dlr (approved):

PR #784, opened 2020-10-12
Update requirements concept to support basic verification

franzTobiasDLR:
Changed requirements trace link to be part of verification. The concept now contains basic elements for generic and domain-specific requirements verification. Other concepts can extend the generic verification interface to create domain-specific requirements validation.

Updates exisiting customized editors to new concept elements...

Otherwise only concept and generated file changes.


de.dlr.sc.virsat.model.extension.requirements.ui/src/de/dlr/sc/virsat/model/extension/requirements/ui/celleditor/DefaultVerificationTypeReferenceCellEditingSupport.java
```diff
@@ -0,0 +1,62 @@
...
+			caDefaultVerification = (CategoryAssignment) element;
+		} else if (element instanceof ComposedPropertyInstance) {
+			caDefaultVerification = ((ComposedPropertyInstance) element).getTypeInstance();
+		} else {
+			return super.createSetCommand(element, userInputValue);
+		}
+		if (userInputValue instanceof CategoryAssignment) {
+			caType = (CategoryAssignment) userInputValue;
+		}
+		if (caDefaultVerification.getType().getFullQualifiedName().equals(DefaultVerification.FULL_QUALIFIED_CATEGORY_NAME)
```
SaMuellerDLR, line 53: There seems to be quite a bit of if branching going on in this method.
Could we maybe move this if block into the above if?
Then we could also move the caType declaration inside and could drop the `caType != null` in the if.

SaMuellerDLR (changes requested):
Looks good. I am not a 100% sure if I caught all the manual changes with all the auto-generated code :D

de.dlr.sc.virsat.model.extension.requirements.ui/src/de/dlr/sc/virsat/model/extension/requirements/ui/celleditor/DefaultVerificationTypeReferenceCellEditingSupport.java
franzTobiasDLR: Not really sure if it's better like this, because we now still have to check if caType is a CA and then also check if it is the right CA... all in one IF... but it's less branching now :D

SaMuellerDLR: Hmm yeah, it looked nicer in my head.
In the future we should maybe consider making a helper function in CategoryAssignmentHelper to check if an object is a ca & has the correct type given by a fqn.
I believe we probably have to check both in a lot of places.

SaMuellerDLR (approved):
Approved!

andriikovalov-dlr (approved):
Looks fine, just some outdated javadocs

Issue #597, opened 2020-05-11
Delete deprecated server classes and tests


JAmmermann-DLR:
E. g. AccessTestResource

PR #800, opened 2020-10-29
Deleted deprecated server classes and tests

JAmmermann-DLR:
Deleted the following classes:
- `Model`
- `VirSatGitAccess` and `VirSatGitAccessTest`
- `AccessTestResource` and `AccessTestResourceTest`
- `WorkspaceAccessResource` and `WorkspaceAccessResourceTest`
- `WorkspaceUserContext` and `WorkspaceUserContextTest`

Also removed now unused git repository from `AGitAndJettyServerTest` and renamed to `AJettyServerTest`

Closes https://github.com/virtualsatellite/VirtualSatellite4-Core/issues/597


de.dlr.sc.virsat.server.test/src/de/dlr/sc/virsat/server/dataaccess/TransactionalJsonProviderTest.java
```diff
@@ -63,7 +63,6 @@
 
 	@Before
 	public void setUp() throws Exception {
-		super.setUp();
```
SaMuellerDLR, line 66: This is a bit strange. If we dont need the setup from the super class, then we are we extending AServerRepositoryTest?

SaMuellerDLR (changes requested):
Very minor detail.

de.dlr.sc.virsat.server.test/src/de/dlr/sc/virsat/server/dataaccess/TransactionalJsonProviderTest.java
JAmmermann-DLR: You are right that is strange. I looked into it and it called the `setUp()` from the `AGitAndJettyServerTest`, that `AServerRepositoryTest `extends, that I deleted in this PR. The problem causing this misunderstanding was wrong naming of the   junit setup/teardown methods in `AServerRepositoryTest`, so I will rename them to make it clear.

SaMuellerDLR (approved):
Approved!

andriikovalov-dlr (approved):

Issue #791, opened 2020-10-19
Add version number to existing REST APIs


JAmmermann-DLR:
The APIs should follow the URL pattern of _http://localhost:port/rest/[API]/[VERSION]/... _.

PR #801, opened 2020-10-29
Added version number to existing REST APIs

JAmmermann-DLR:
Now the APIs follow the URL pattern of _http://localhost:port/rest/[API]/[VERSION]/... _.

Each Servlet now has its own name and version number, that is registered in the `VirSatJettyServer`

Fixed the API paths in the test cases

Closes https://github.com/virtualsatellite/VirtualSatellite4-Core/issues/791


SaMuellerDLR (approved):
Approved!

PhilMFischer (changes requested):
At least i would like to see, that we have a follow on ticket, to activate authentication for the upcoming release. At minimum user/password by file configuration should be turned on as standard.

JAmmermann-DLR:
> 
> 
> At least i would like to see, that we have a follow on ticket, to activate authentication for the upcoming release. At minimum user/password by file configuration should be turned on as standard.

Created follow up ticket https://github.com/virtualsatellite/VirtualSatellite4-Core/issues/812

PhilMFischer (approved):

Issue #802, opened 2020-10-29
Fix abstract BeanProperty classes inconsistency


JAmmermann-DLR:
The exiting abstract BeanProperty classes do not reflect the structure of the underlying type instances correctly. This inconsistency causes problems, e.g. no abstract class for all bean properties exists.

- `ABeanProperty` only wraps `ValuePropertyInstance` thus it should be renamed to `ABeanValueProperty`
- `ABeanUnitProperty `wraps the `UnitValuePropertyInstance `thus should be renamed to `ABeanUnitValueProperty`
- A new  `ABeanProperty` should be added that wraps `APropertyInstance` and is extende by all bean property classes

PR #803, opened 2020-10-30
Fixed abstract BeanProperty classes inconsistency

JAmmermann-DLR:
- Renamed `ABeanProperty` to `ABeanValueProperty`
- Renamed `ABeanUnitProperty` to `ABeanUnitValueProperty`
- Added  `ABeanProperty` that wraps `APropertyInstance` and is extended by all bean property classes
- All bean constructores now use `super` instead of `setTypeInstance`
- Removed redundand `implements` statements

Closes https://github.com/virtualsatellite/VirtualSatellite4-Core/issues/802


SaMuellerDLR (approved):
Approved!

PhilMFischer (approved):
Looks all good to me. I cannot oversee all side effects, but seems to make sense to follow the object hierarchy for property instances that we also have on ecore level. I also like the new semantic typing which is indirectly introduced to state that a bean is a PropertyInstance.

Issue #816, opened 2020-11-05
NPE in graphiti when opening a diagram after creating a connection


franzTobiasDLR:
When opening a graphiti editor and there is a Add\*ConnectionCommand in our command stack, then the initalisation of the editor will initialize undo/redo actions by exectuing the canExecute of this last command. As we do not have a graphical diagram part at this point of the editor initialization, the undo of the Add\*ConnectionCommand will result in a NPE.

PR #817, opened 2020-11-05
Fix NPE when opening a graphiti diagram

franzTobiasDLR:
- Make sure the graphical viewer has an editor control when calculating
the real mouse location
- If no control exists yet because the editor is being initialized, then
simply return the native location to initialize the editor actions

Closing #816


SaMuellerDLR (approved):
Approved!

GeorgeNasrDLR (approved):
Done!

PR #819, opened 2020-11-06
Updated state machine concept - (Task #818)

### LINKED ISSUE #818: Extend state machines concept with additional properties
- Add long description propery 
- Add trigger property

SaMuellerDLR:
- Added simple detail description field for states
- Added a category to decribe events that trigger transitions

Closes #818

Transition trigger events can be customized by specialised implementations. We could for example have a telecommand trigger or a failure trigger...

Diagram label providers can be registered via extension registry to enable domain specific transition labels such as TC with its paramters....

---
Task #818: Extend state machines concept with additional properties


franzTobiasDLR (approved):

de.dlr.sc.virsat.model.extension.statemachines/concept/concept.concept
```diff
@@ -7,15 +7,17 @@
...
 	Import de.dlr.sc.virsat.model.extension.ps.*; 
 	
 	Category State description "State of the component at a given time"{
+		StringProperty detail description "Optional additional detail information about the state";
 	}	
 	
 	Category Transition description "Transitions among states" { 
 		Reference stateFrom of Type State description "The Transition starts in this state";
 		Reference stateTo of Type State description "The Transition ends in this state";
+		Reference trigger of Type GenericCategory description "The event that triggers the Transition";
```
PhilMFischer, line 20: SHould a trigger be of type generic? I think we are sacrifying some semantics for the sake of simplicity. I think we should at least provide a simple Category such as a generic Event or Trigger that can be referenced and used in other Concepts.

PhilMFischer (changes requested):
I think we should reconsider the change to the concept to improve to enhance the semantics.

PhilMFischer (changes requested):
I think loading the labelprovider from the platform registry should be reviewed and most likely be improved.

PhilMFischer (approved):

PR #857, opened 2020-11-30
Update the icons of CostEquipment and CostSummary

kempra:
Updated the Icons of the Categories (CostEquipment,CostSummary).


SaMuellerDLR (approved):
Approved!

Issue #866, opened 2020-12-01
Restore the Concep.concept in the Cost-Concept


kempra:
Deleted all autogenerated Files that are not used.

PR #867, opened 2020-12-01
Restore the Concep.concept in the Cost-Concept - (Task #866)

kempra:
Restore the Concep.concept in the Cost-Conzept.
Delete all auto generated files that are no longer in use.


SaMuellerDLR (approved):
Approved!

PR #870, opened 2020-12-02
Added CorsFilter to handle CORS requests

### LINKED ISSUE #826: Add CORS Headers to Server Response
The Server should respond with the correct CORS Headers in the Response to CORS Requests, else APIs calling the Server via CORS will receive errors. This behaviour could be achieved using a Request/Response Filter.

JAmmermann-DLR:
The filter appends the required headers for the CORS Requests. This feature is needed to call the Server API e.g. from JavaScript running in the Browser. 

Closes #826


JAmmermann-DLR:
I think the problem with the current approach on Jersey Servlet level is some caching in the HttpURLConnection on the system properties.
Possible solutions for this problem are:
- Directly start the Java VM of the with this 
- Set the system property somewhere before the `HttpURLConnection `is created
- Use another connector that doesn't use the `HttpURLConnection `

When I read more about the connectos, I realized that we should probably approach that issue serverwide on the Jetty level instead.
There should be a `org.eclipse.jetty.servlets.CrossOriginFilter` available, working with our Jetty version, so I will look into this one next.

de.dlr.sc.virsat.server/src/de/dlr/sc/virsat/server/auth/filter/CorsFilter.java
```diff
@@ -0,0 +1,78 @@
...
+public class CorsFilter implements ContainerRequestFilter, ContainerResponseFilter {
+
+	public static final String AC_ALLOW_CREDENTIALS = "Access-Control-Allow-Credentials";
+	public static final String AC_ALLOW_METHODS = "Access-Control-Allow-Methods";
+	public static final String AC_ALLOW_HEADERS = "Access-Control-Allow-Headers";
+	public static final String AC_ALLOW_ORIGIN = "Access-Control-Allow-Origin";
+	
+	public static final String AC_ALLOWED_CREDENTIALS = "true";
+	public static final String AC_ALLOWED_METHODS = "GET, POST, PUT, DELETE, OPTIONS, HEAD";
+	public static final String AC_ALLOWED_HEADERS = ""; //"X-Requested-With, Authorization, Accept-Version, Content-MD5, CSRF-Token, Content-Type";
```
SaMuellerDLR, line 36: What is the purpose of the comment here?

SaMuellerDLR (approved):
Great work! Only a very minor unclarity regarding the purpose of one of the comments.

de.dlr.sc.virsat.server/src/de/dlr/sc/virsat/server/auth/filter/CorsFilter.java
JAmmermann-DLR: Ok so these are headers that we possible could/need to support at some point. I don't know if it's actually smart to leave them there or if we should delete them for now. I have no strong opinion on it.

franzTobiasDLR (approved):

PR #871, opened 2020-12-02
Fixed Problems on GitHub Actions build on Ubuntu 18.04 nodes

### LINKED ISSUE #846: GitHub Actions - Bionic Migration
Change GitHub Actions to use Ubuntu 18.04 instead of 16.04 computers.

ghost:
Fixed problems on Ubuntu 18.04 machines with failing builds.

- Fixed an issue with DmfResource tests which was due to having two times the same test class in the same package in different plugins.
- Fixed swt bot test case for editor which was indirectly issueing "enter" key strokes.
- Fixed setup shell scripts to improve detection of dependencies such as vtk and zmq.
- Fixed setup shell scripts to correctly set localisation support for ubuntu 18.04 build nodes. (Fixed AwtKeyboardStrategy exception in editor testc ases)
- Removed openssl 1.1.1 installation since it is already included in 18.04


de.dlr.sc.virsat.project/src/de/dlr/sc/virsat/project/resources/dmf/DmfResource.java
```diff
@@ -98,7 +98,7 @@ protected void saveDMFResource(Map<?, ?> options) throws IOException {
 		// Use NoUndo to suppress notification of this infrastructure-related
 		// operation
 		virSatEd.getVirSatCommandStack().executeNoUndo(dmfResourceSaveCommand);
-		virSatDvlmResource.save(options);
+		virSatEd.saveAll();
```
franzTobiasDLR, line 101: I kind of remember that we changed something like that in the early times of DMF because of ti was messing up the order of sving resources, let me check that it is still working as we expect it.

PhilMFischer: Yes would be great if you could check this. We had two DmfResource Test classes which was creating trouble for whatever reason. Nevertheless saving a dvlm resource outside the editing domain sounds very error prone. If there is an issue with the order of saving we should carefully review what we are doing here.

maven/pom.xml
```diff
@@ -184,6 +184,7 @@
 							<redirectTestOutputToFile>true</redirectTestOutputToFile>
 							<showEclipseLog>false</showEclipseLog>
 							<testFailureIgnore>false</testFailureIgnore>
+							<failIfNoTests>false</failIfNoTests>
```
SaMuellerDLR, line 187: Thats a cool flag. If we have this, do we still need the dummy template test?

SaMuellerDLR (changes requested):
Only a minor question.

maven/pom.xml
PhilMFischer: Unfortunately yes, because it just avoids a fail but still throws a warning, which makes the maven_build script fail. Of course we could add the warning to the exceptions, but from a maven point of view, with convention over configuration I prefer the template test for the moment.

SaMuellerDLR: Alright!

SaMuellerDLR (approved):
Approved!

de.dlr.sc.virsat.project/src/de/dlr/sc/virsat/project/resources/dmf/DmfResource.java
franzTobiasDLR: After some intence testing I didn't find any problems.... Let's try it.

franzTobiasDLR (approved):

PR #888, opened 2020-12-04
Improved build stability 

SaMuellerDLR:
- Re-added downloading and installing of Overtarget into Maven m2. Gost lost in a previous PR and only worked due to m2 caching.
- Changed to different version of DmfResourceTest (With a Class absed rule for timeout but without inheriting from AProjectTestCase which has a global timeout rule and causes jacoco issues on random)
- CHanged acquision of Problems view in RoleManagementTest to get it by ID rather than name since there are known issues from doing that in ValidatorTests.

### DIFF
--- a/bash/setup.sh
+++ b/bash/setup.sh
@@ -22,7 +22,7 @@ sudo apt-get install openjdk-8-jdk xvfb metacity libvtk6-java libzmq5 libzmq-jav
 # cd ..
   
 # -------------------------------------
-# Installing OverTarget
+# Installing Overtarget
 # -------------------------------------
 
 echo "Installing and running Overtarget"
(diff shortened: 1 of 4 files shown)


bash/setup.sh
```diff
@@ -21,6 +21,13 @@ sudo apt-get install openjdk-8-jdk xvfb metacity libvtk6-java libzmq5 libzmq-jav
 # sudo make install
 # cd ..
   
+# -------------------------------------
+# Installing OverTarget
```
Anna-Lena-DLR, line 25: Just a really small thing: we decided to write Overtarget instead of OverTarget.

bash/setup.sh
```diff
@@ -21,6 +21,13 @@ sudo apt-get install openjdk-8-jdk xvfb metacity libvtk6-java libzmq5 libzmq-jav
...
+# -------------------------------------
+
+echo "Installing and running OverTarget"
```
Anna-Lena-DLR, line 28: see above

Anna-Lena-DLR (approved):
Looks good

franzTobiasDLR (approved):
Coverage issues again...  🙈

PR #901, opened 2020-12-09
Hide composed property instance 

franzTobiasDLR:
Forward content provider of composed property instance to its type instance to make the navigator cleaner...

Closing #901


andriikovalov-dlr (approved):

SaMuellerDLR (approved):
Approved!

PR #910, opened 2021-01-12
Feature/task 845 Add a JFreeChart Visualization for the Cost Concept

kempra:
* Adds JFreeChart piechart and stack chart similar to the mass concept to visualize the budget distribution of the costs
* Handled case of costs without assigned type


SaMuellerDLR (approved):
Approved!

maurpa (approved):

PR #924, opened 2021-01-25
Fixed deployed tar file generation for fixing executable permissions

### LINKED ISSUE #921: Make VirSat executable in Linux
In Linux VirSat executable file does not have executable flag. Would be nice to mark it executable, so that the user doesn't need to do it manually.

ghost:
- Fixed ant step for constructing deployable ant files to set executable permissions on VirSat Core and Server executable
- Fixed memory settings in launcher to avoid out of memory exception
- Added build step to github actions workflow for uploading all deployable zips and tar.gzs.

Closes #921

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -288,6 +288,7 @@ jobs:
           path: |
             deploy/**/*.zip
             deploy/**/*.tar.gz
+          retention-days: 3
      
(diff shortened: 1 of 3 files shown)


andriikovalov-dlr (approved):

SaMuellerDLR (approved):
Approved!

PR #927, opened 2021-01-26
Made bean SEI use reference instead of deep nesting

### LINKED ISSUE #923: Make server beans nested by reference only
Instead of deep nested beans only hand out shallow beans with references.

JAmmermann-DLR:
Instead of nested unmarshalling  now the bean seis only have a reference to other bean seis. The first idea was to just use the BeanStructuralElementInstance similar to the way we use the BeanCategoryAssignment, but there was no way to exclude fields of the inherited ABeanStructuralElementInstance (e.g. children), because of the XML propOrder that we have to use there. So instead I resolved the problem the following way:
- Added BeanStructuralElementInstanceReference that only has the fields name and uuid
- Use it in ABeanStructuralElementInstance in the jaxb functions so the interface is not changed
- Adjusted the test json accordingly
- Fix a bug in the documentation, so that the parent bean of a sei is correctly displayed as uuid string (by using the model substitution)
- Regenerate the swagger documentation

Closes #923


SaMuellerDLR (approved):
Approved!

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/types/structural/ABeanStructuralElementInstance.java
```diff
@@ -219,20 +219,49 @@ public void setChildren(List<ABeanStructuralElementInstance> newBeanSeis) {
...
+	public List<BeanStructuralElementInstanceReference> getJaxbChildren() {
+		return getReferenceList(sei.getChildren());
 	}
 	
-	//CHECKSTYLE:OFF: MagicNumber
 	@XmlElement(name = "children")
+	//CHECKSTYLE:OFF: MagicNumber
 	@ApiModelProperty(name = "children", required = true,
 		value = "List of the child beans",
 		position = 3)
```
franzTobiasDLR, line 261: Why do we need this magic number here?

franzTobiasDLR (approved):

de.dlr.sc.virsat.model.edit/src/de/dlr/sc/virsat/model/concept/types/structural/ABeanStructuralElementInstance.java
JAmmermann-DLR: Because of the `position = 3`. Without the editor shows a magic number warning. (tested it right now in my IDE to be sure)

JAmmermann-DLR: The alternative would be to use constants

JAmmermann-DLR: Also before there were only two CHECKSTYLE comments for the whole file, I now moved them around the problematic annotations to improve code quality. Maybe thats what confused you

franzTobiasDLR: Yep, I know that checkstyle complaints about the magic numbers... That's what it is for :D My question was more like can't we use Constants here?

JAmmermann-DLR: Sure I will change it then

PR #935, opened 2021-02-02
Added Server secure communication capabilities

### LINKED ISSUE #531: Server secure communication capabilities
When communicating with the server passwords for the repository have to be transmitted. therefore the server needs to have the capability to communicate also via _https:_. The certificates have to be stored in the configuration area (#530) .  The virtual satellite server does not provide pre generated certificates stored in the repository (_No shipping of certificates_) .  But the user interface allows to simply create them. Add functionality to disable http communication with the jetty server. 

- [x] implement core https functinality
- [ ] implement simple certificate wizzard from command line
- [x] add setting to disable http access

JAmmermann-DLR:
- Added HTTPS capabilities
  - Added new fields for HTTPS to the `ServerConfiguration` and properties files
  - Added HTTPS to `VirSatJettyServer` that uses an existing keystore that has to be provided via the properties
  - Added a test case for the `VirSatJettyServer`, that tests HTTP / HTTPS
  - Added a status servlet that is used by the test above and allows the user to test server connection without authentication
- Added HTTPS to server documentation

- Did NOT add certificate wizzard from command line, it should be discussed if we really want to support this feature and possibly a new ticket should be created for that issue

Closes #531


SaMuellerDLR (approved):
Approved!

maurpa (approved):

PR #946, opened 2021-02-17
Added missing adapter to TransactionalJsonProvider - (Task #945)

### LINKED ISSUE #945: Fix nullpointer on PUT of Sei via server
Currently throws a NPE because an Adapter is missing the resource set.

JAmmermann-DLR:
This fixes the NPE.
Also added a test case.

Closes #945: Fix nullpointer on PUT of Sei via server


maurpa (approved):

SaMuellerDLR (approved):
Approved!

PR #951, opened 2021-02-22
Added delete endpoints to server

### LINKED ISSUE #949: Add delete endpoints to server
Add an endpoint to remove SEIs or CAs via the REST API.

JAmmermann-DLR:
Added delete endpoints to the REST API for CAs and SEIs.

Closes #949

### DIFF
--- a/de.dlr.sc.virsat.server.test/src/de/dlr/sc/virsat/server/resources/ModelAccessResourceTest.java
+++ b/de.dlr.sc.virsat.server.test/src/de/dlr/sc/virsat/server/resources/ModelAccessResourceTest.java
@@ -10,6 +10,7 @@
 package de.dlr.sc.virsat.server.resources;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
@@ -25,6 +26,7 @@ import javax.ws.rs.core.MultivaluedHashMap;
 import javax.ws.rs.core.Response;
 import javax.xml.bind.JAXBException;
 
+import org.eclipse.core.resources.IFolder;
 import org.eclipse.emf.transaction.RecordingCommand;
 import org.eclipse.jetty.http.HttpStatus;
 import org.eclipse.jgit.api.Git;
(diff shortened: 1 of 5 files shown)


JAmmermann-DLR:
Following points have to be discussed:
- Should deleting a SEI via the server also remove its folder on the file system?
  - Deleting without file system could be easily implemented similar to the CA deletion. 
  - Probably we could also provide both options (deletion on fs and only in the project) via an URI param if that makes sense.
- Should there be an endpoint for deleting properties? I don't think we can delete them without the ca, right? Does this also hold for dynamic bean list elements?
- Another problem I see now is that the resource class itself (and its test class) is getting pretty big. And as we are know moved synchronization in that class, should we introduce an controller class for this resource (in another ticket)?

SaMuellerDLR:
> Following points have to be discussed:
>     * Should deleting a SEI via the server also remove its folder on the file system?
>       * Deleting without file system could be easily implemented similar to the CA deletion.
>       * Probably we could also provide both options (deletion on fs and only in the project) via an URI param if that makes sense.
I think ideally the server delete should work the same as when deleting in the desktop application, so by default it should also be deleted on the filesystem. But I see you are already using CreateRemoveSeiWithFilteStructure command which does exactly that, no?
>     * Should there be an endpoint for deleting properties? I don't think we can delete them without the ca, right? Does this also hold for dynamic bean list elements?
In general, properties (or more precise property instances) cannot be deleted. I think the only exception are property instance listst, But I am not 100% sure who they are implemented.
>     * Another problem I see now is that the resource class itself (and its test class) is getting pretty big. And as we are know moved synchronization in that class, should we introduce an controller class for this resource (in another ticket)?
Would it maybe make sense to split the class up into SeiResourceAccess / CaResourceAccess / PiResourceAccess / etc.?

JAmmermann-DLR:
> 
> 
> > Following points have to be discussed:
> > * Should deleting a SEI via the server also remove its folder on the file system?
> > * Deleting without file system could be easily implemented similar to the CA deletion.
> > * Probably we could also provide both options (deletion on fs and only in the project) via an URI param if that makes sense.
> 
> I think ideally the server delete should work the same as when deleting in the desktop application, so by default it should also be deleted on the filesystem. But I see you are already using CreateRemoveSeiWithFilteStructure command which does exactly that, no?

Yes that is right and that makes sense to me. 
One further question regarding this topic: I was wondering why the `ABeanStructuralElementInstance` implementation of the delete method then does not delete on the file system and instead only uses the `DeleteStructuralElementInstanceCommand`. Maybe we could change the call there and then only call this function in the endpoint?
> 
> > ```
> > * Should there be an endpoint for deleting properties? I don't think we can delete them without the ca, right? Does this also hold for dynamic bean list elements?
> > ```
> 
> In general, properties (or more precise property instances) cannot be deleted. I think the only exception are property instance listst, But I am not 100% sure who they are implemented.

Ok then I would propose that we ignore it for now and deal with it when we fix the lists.
> 
> > ```
> > * Another problem I see now is that the resource class itself (and its test class) is getting pretty big. And as we are know moved synchronization in that class, should we introduce an controller class for this resource (in another ticket)?
> > ```
> 
> Would it maybe make sense to split the class up into SeiResourceAccess / CaResourceAccess / PiResourceAccess / etc.?

Yeah that sounds good. Then lets try that before continuing with the Create endpoints, so that we don't have so much to refactor?

SaMuellerDLR (approved):
Approved!

maurpa (approved):

PR #969, opened 2021-03-19
Add unique identifier to concrete bean JSONs

### LINKED ISSUE #962: Add unique identifier to concrete bean JSONs
A minimal example has shown that un-/marshalling Classes with the same name in different packages result in a _Name collision_.
This problem can be solved by adding the `@XmlType(name=...)` annotation to the concrete classes.
In the beans the constant of the FQN from the abstract bean should be used  with that annotation (e.g. for `TestStructuralElement` add `@XmlType(name = ATestStructuralElement.FULL_QUALIFIED_STRUCTURAL_ELEMENT_NAME)`. This has to be changed for all generated beans via the code generators.

Note: Additional to solving the uniqueness issue, this change also results in consistent naming, e.g. when creating a new bean the same FQN has to be provided via the REST enpoint, as it is used as identifier in the `BeanRegistry`.

JAmmermann-DLR:
- Add unique identifiers to the concrete beans using `@XmlType`
- Add this annotation also to the generators, for new beans in the future
- Refactor the adapter for composed beans, because it somehow needs stronger typing now to work correctly...

Closes #962


maurpa (approved):

SaMuellerDLR (approved):
Approved!

PR #971, opened 2021-03-21
Task 967 add cost type import wiz

kempra:
* Implements an import wizard for importing (or reimporting) cost budgests from excel tables
* Implementeded test cases
* Added a user manual chapter for using modeling cost budgets


SaMuellerDLR (approved):
Approved!

maurpa (approved):

PR #976, opened 2021-04-07
Expose EditingSupport of AUiSnippetArrayInstancePropertyTable

watollah-dlr:
For AUiSnippetGenericCategoryAssignmentTable it is possible to provide a custom EditingSupport to modify the editing in the table columns. This should be possible for AUiSnippetArrayInstancePropertyTable as well.

I suggest we just move the method createEditingSupport from AUiSnippetGenericCategoryAssignmentTable to the shared parent class AUiSnippetGenericTable.


GeorgeNasrDLR (approved):

SaMuellerDLR (approved):
Approved! (Please merge with development to get rid of error messages!)

PR #984, opened 2021-04-26
Added create endpoints

JAmmermann-DLR:
Added three create endpoints and test cases for:
- RootSei in ModelAccessResource
- Sei in StructuralElementInstanceResource
- Ca in CategoryAssignmentResource

Also added some minor improvement in the server errors.

@SaMuellerDLR the only think we don't handle at the moment are composed beans. For example what if one wants to create a ca inside a composed bean?

Closes #964

PR #985, opened 2021-04-26
Added query param no sync

JAmmermann-DLR:
- Added a query param to toggle synchronization and builds on ALL model requests
- Added endpoint to force synchronization (ignoring the query parameter)
- Minor refactoring in `ServerRepository` and subresources

To test the synchronization from the REST enpoints is not so easy. My idea was to just check the log messages, but I am not sure if that is the way to go. The alternative would be to for example temper with some files in the git and afterwards check that they updated.

Closes #984

### DIFF
--- a/de.dlr.sc.virsat.server.test/src/de/dlr/sc/virsat/server/resources/AModelAccessResourceTest.java
+++ b/de.dlr.sc.virsat.server.test/src/de/dlr/sc/virsat/server/resources/AModelAccessResourceTest.java
@@ -9,8 +9,6 @@
  *******************************************************************************/
 package de.dlr.sc.virsat.server.resources;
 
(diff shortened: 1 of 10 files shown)


JAmmermann-DLR:
I changed the build type to incremental, which worked on my local maschine, but not in the CI. (See https://github.com/virtualsatellite/VirtualSatellite4-Core/pull/985/commits/dc85eead2e3b19c785ce556da9daf546b64e13b9) I assume that the incremental build on the CI did not remove the dangling reference (and thus did not create an additional commit). Has anyone an idea why it behaves that way? If we want to switch to an incremental build, we should probably open a new ticket to investigate it further.

SaMuellerDLR (approved):
Approved!

franzTobiasDLR (approved):

Issue #986, opened 2021-04-26
Add requirement test resources to build configuration


franzTobiasDLR:
Add the ReqIF test resources to the build configuration so that they can be used in downstream proejcts

PR #987, opened 2021-04-26
Add ReqIF test files to build includes

franzTobiasDLR:
Add ReqIF fieles to buidl config so that they can be used from downstream projects?!

Closes Task #986


SaMuellerDLR (approved):
Approved!

maurpa (approved):

PR #994, opened 2021-06-02
PR1: Added qudv beans

JAmmermann-DLR:
This is the first PR of a number of PRs to enable unit management via the server (see #980). This PR adds beans for qudv classes.

Added:
- Beans for units and quantity kinds
- Beans for factors and prefix
- Bean factories for unit and quantity kind beans
- Unit beans to existing unit property beans

Did not add beans for `Dimension`, `DimensionFactor` and `QuantityKind`, as I found no use case for these model Classes.

In a second PR these beans should be annotated and made available via the REST API.


SaMuellerDLR (approved):
Approved!

maurpa (approved):

Issue #78, opened 2019-06-19
Create a performCalculations method in ModelAPI


SaMuellerDLR:
Its possible to perform an inheritance update through the ModelAPI via the method performInheritance(). This method should exist also for updating calculations.

PR #1002, opened 2021-06-24
Feature/task 1000 change performance tests

### LINKED ISSUE #1000: Refactor performance tests

andriikovalov-dlr:
closes #1000

I didn't manage to add calculations because modelAPI doesn't support it yet, and running the equations builder on the test project didn't work for me. We should revisit it after we do #78.


SaMuellerDLR (approved):
Approved!

GeorgeNasrDLR (approved):

PR #1006, opened 2021-06-30
PR2: Added annotations to qudv beans

JAmmermann-DLR:
Follow up PR for #994 (see #980)

Added:
- Annotations to the beans
- Adapters for reference relations
- Test cases
- Had to change some Interfaces to Abstract Class for the un-/marshalling to work

In a final PR the annotated beans can be made available via the REST API.


SaMuellerDLR (approved):
Approved!

maurpa (approved):

PR #1007, opened 2021-07-01
Added is calculate method for bean properties

### LINKED ISSUE #997: Handling of computed properties in the Server
Computed properties are not handled by the server yet. Users of the API should at least know which property is computed and probably not be able to change them directly.

### LINKED ISSUE #1005: Add isCalculated field to bean properties
So that clients know which proeprties are calculated and thus should not be changed via the API.

JAmmermann-DLR:
Add special handling for composed beans.
Update test resources.

Closes #1005
Closes #997 (duplicate)


GeorgeNasrDLR (approved):

SaMuellerDLR (approved):
Approved!

PR #1011, opened 2021-07-27
Feature/task 1004 add unit setters and getters to bean property float

PeterOhrDLR:
In this branch two new methods are added to the `BeanPropertyFloat.java` class. The purpose of the methods is to be able to get correct values with respect to the default unit of a property or with respect to a unit passed as parameter. The default unit is specified in the concept which contains the respective category.
Furthermore Peter Ohr was added as an author.


SaMuellerDLR (approved):
Approved!

andriikovalov-dlr (approved):

PR #1081, opened 2022-03-31
Performance Imprtovements from Active Concept Helper

ghost:
The Active Conept Helper is used to calculate IDs and FQNs in particular for various objects. The calculation of FQN is quite costly and can be cached, since they are calculated from the types in the concept, which only change by migration. The ActiveConceptHelper is adjusted accrodingly to cache these FQN rather than recalculating them all the time. The adjustments also concerns:

- The integration into the DvlmXMIReosurceSet.
- The invalidation/clearing of the cache.
- A change in the behaviour of the getFQN methon in the IQualifedName interface, which now does not return the noid default string anymore, but a null in case the FQN contains the noid string.


SaMuellerDLR (approved):
Approved!

franzTobiasDLR (approved):

Issue #743, opened 2020-10-05
Create Doors Requirements Synchronization


Anna-Lena-DLR:
Implement http connection to Doors NG to automatically synchronize requirements. Use OSLC specification.

PR #1082, opened 2022-04-05
WIP: Feature/task 743 add external lib

Anna-Lena-DLR:
Added required external libs for Doors synchronization process

Linked to #743


de.dlr.sc.virsat.external.lib.lyo/about_files/NOTICE
```diff
@@ -0,0 +1,23 @@
+Apache POI
```
PhilMFischer, line 1: This is not Apache POI

de.dlr.sc.virsat.external.lib.lyo/about_files/THIRD-PARTY.txt
```diff
@@ -0,0 +1,66 @@
+
+Lists of 64 third-party dependencies.
+     (Apache 2.0) Jackson-annotations (com.fasterxml.jackson.core:jackson-annotations:2.10.1 - http://github.com/FasterXML/jackson)
+     (Apache 2.0) Jackson-core (com.fasterxml.jackson.core:jackson-core:2.10.1 - https://github.com/FasterXML/jackson-core)
+     (Apache 2.0) jackson-databind (com.fasterxml.jackson.core:jackson-databind:2.10.1 - http://github.com/FasterXML/jackson)
```
PhilMFischer, line 5: Jackson dependencies are already introduced via the external lib jersey. should check if we need them here again, or if we have to externalize it.

de.dlr.sc.virsat.external.lib.lyo/about.html
```diff
@@ -0,0 +1,57 @@
...
+for informational purposes only, and you should look to the Redistributor&rsquo;s license for 
+terms and conditions of use. Third party licenses are aggregated in the &quot;about_files&quot; folder of this plugin.
+</p>
+
+<h4>Directly Maven Included Thrid Party Libraries</h4>
+
+The following third party libraries are directly included into this plugin using maven. 
+Transitive dependency information is given at the end of the page.
+
+<h5>Apache POI - Version 3.17</h5>
```
PhilMFischer, line 40: Apache POI should not be mentioned here. Please update the descriptions of the dependencies.

de.dlr.sc.virsat.external.lib.lyo/externalLib/maven/apache-jena-libs.pom
```diff
@@ -0,0 +1,67 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>apache-jena-libs</artifactId>
```
PhilMFischer, line 23: In case we need jena, it should be considered to be an external library by itself.

de.dlr.sc.virsat.external.lib.mockito/about.html
```diff
@@ -0,0 +1,57 @@
...
+for informational purposes only, and you should look to the Redistributor&rsquo;s license for 
+terms and conditions of use. Third party licenses are aggregated in the &quot;about_files&quot; folder of this plugin.
+</p>
+
+<h4>Directly Maven Included Thrid Party Libraries</h4>
+
+The following third party libraries are directly included into this plugin using maven. 
+Transitive dependency information is given at the end of the page.
+
+<h5>Apache POI - Version 3.17</h5>
```
PhilMFischer, line 40: This one should state something about Mockito

ghost (changes requested):
I think, this may become a highly complicated integration.

I suggest that we do the Java 11 and Eclipse 2022-03 migration first, before going back to this PR.

I see the point of the external lib for LYO, but i think we shoiuld be careful with choosing dependencies, since a lot of dependencies are already provided by our external libs for the jersey integration. We may need to clean up here, otherwise we might end up in running to many different versions in parralel.

Another point is the use of Jena which is not clear to me. I also question if it needs to be separated into an individuel external lib plugin.

The About.htmls are not correctly adjusted in all cases. They all state a third paryt dependency to Apache POI, this has to be cleaned up.

I suggest, that we discuss the integration in one of our next developer meetings, so that everyone is aware of these upcoming changes.

PR #1089, opened 2022-05-09
Changed VirSatNavigatorSeiSorter to return zero - Task #1088

### LINKED ISSUE #1088: Change VirSatNavigatorSeiSorter to maintain the order given by content provider
Currently `VirSatNavigatorSeiSorter` sorts items based on the indices in the containing lists, and for each comparison calls `list.indexOf()`, which is itself linear, making the whole sorting complexity square*log.

We could instead just return zeros, and maintain the order given by the content provider, which is presumably also the order of containing lists.

andriikovalov-dlr:
Tested on KatSat, seems to be the same order (checked top level and some random nested components)

Closes #1088


SaMuellerDLR (approved):
Nice simplification! Approved!

PR #1115, opened 2022-07-11
Integration 4.15.0 - Remerge versions

SaMuellerDLR:
- Updated version to 4.25.0
- Updated DVLM version to v9 and regenerated all concepts
- Regenerated XText languages and Ecore models
- Fixed a bug in the generator that prevented the core concept from being regenerated upon DVLM version increase


GeorgeNasrDLR (approved):

GeorgeNasrDLR:
Done!

PR #1125, opened 2023-04-26
Enable creation of self-loops in state machines

pchrszon-dlr:
- When creating transitions using the state-machine diagram editor, a state can now be select both as source state and as target state of a transition (by clicking on the same state twice), resulting in a self-loop.
- Bend points are automatically created on a loop transition to improve visibility and usability.


franzTobiasDLR (approved):
Looking good!

dellerDLR (approved):

dellerDLR:
Was merged in #1154

PR #1148, opened 2023-08-16
Feature/1146 change default drag drop behavior for product structures

### LINKED ISSUE #1146: Change default drag & drop behavior for product structures
Currently, dragging an element from, e.g., the product tree to the configuration tree creates an inheritance link. Conversely, dragging in combination with the ALT-key (i.e., dragging a link) results in a newly created element instance with the correct inheritance link in place. This is a bit counterintuitive and perhaps the opposite of the expected behavior. Rather, dragging a link should create the inheritance link, and dragging without modifiers should create a new instance.

pchrszon-dlr:
This change swaps the standard drag-and-drop behavior and the modified behavior.

Dragging an element from one tree to another now creates a new instance that inherits from the dragged element. The creation of inheritance links is now accomplished by dragging and dropping while holding the ALT-key.

Closes #1146


dellerDLR (approved):
Looks good to me and works as intended.

franzTobiasDLR:
> @pchrszon-dlr @franzTobiasDLR Just some formality: I noticed that the commits belonging to this pull request don't include the Issuenumber (#1146 ) like in the previous pull requests.
> 
> I noticed that there are also commits in the development branch without a issue number, if there was not a issue created for them. How do we handle that?

That's not really a problem. In future we should follow the pattern but I wouldn't change all the commit messages now afterwards...

franzTobiasDLR (approved):

PR #1183, opened 2024-03-28
Unpredicted Close of Editors

ghost:
Removed nested async job in display thread.

This was leading to the problem, that the actual close of the editor could be placed behind a wait for ui therad in the SWTBot tests.

The async job execution is now happening in the super method of the FormEditor.


franzTobiasDLR (approved):
Nice, thank you!

dellerDLR (approved):
Looks good to me!

PR #1193, opened 2024-08-06
Integration 4.17.0 - Remerge Versions

PhilMFischer:
- Updated main version number to 4.17.0
- Created DVLM release under namespace v11
- Regenerated Xtext languages and Concepts with namespace v11

⚠️ Due to complexity of the upgrade, the change of versions is done on a separate branch. Therefore this "Remerge Versions" is happening from feature/task_1192-Release_Version_4.17-Update_Versions . The integration branch will be started from here before merge conflicts to the development branch are resolved. ⚠️


PhilMFischer:
Will not work to remerge versions only. Have to do a final merge with all the migrations

Issue #1218, opened 2024-10-17
Add capability to scale objects in visualization


PhilMFischer:
In case stl files are displayed they are drawn in meters even though the raw data sais centimeters. This results in 100 times bigger drawings. It should therefore be possible to set a scale in the visulaization category that is correctly interpreted when rendering the stl file.

PR #1221, opened 2024-10-22
Added capability to scale STL geometries in the visualization

ghost:
- Added a unit factor to the visualization category. It is supposed to represent want one meassurement in an STL should mean. It is set now to 1mm by default.

- Also adjusted the protobuf scenegraph and the conept as well as the vtk visualization


pchrszon-dlr (approved):

kain-ap (approved):

PhilMFischer:
Closes #1218

PR #1229, opened 2024-10-30
Force build before committing (Task #85)

### LINKED ISSUE #85: Committing without saving can create out-of-date data
When performing an update to some data and then not saving, it can create out of date data in the commited repository.
To reproduce (in CEF, but a CORE problem):
- Create a new VirSat Project
- Use Templates -> Create DLR/CEF Study to create some data
- Share the project into an SVN repo
- Open an equipment in the editor
- Change the massPerUnit parameter
- DO NOT SAVE
- Commit
- Check the data in the repo: It will say have the old value in the calculated massTotalWithMargin and massTotal parameters

This is because commiting causes a Save but does not cause a build!

The problem can be migitated by either:
* Also triggering a build before commiting (possibly with a dialogue asking for confirmation)
* Dont allow to commit while the project hasnt been saved

pchrszon-dlr:
Add a forced build between save-all and commit (both for SVN and git).

It may be the case (still unconfirmed) that after the automatic save-all triggered by the commit handler, the project builders (calculations, inheritance) are triggered and modify files while the commit is in progress.

This change introduces a mandatory build step after the save-all to ensure all builders have been run before the commit. It also disables auto-build while the commit is in progress.

### DIFF
--- a/de.dlr.sc.virsat.team.ui/src/de/dlr/sc/virsat/team/ui/util/VersionControlJob.java
+++ b/de.dlr.sc.virsat.team.ui/src/de/dlr/sc/virsat/team/ui/util/VersionControlJob.java
@@ -14,6 +14,9 @@ import java.util.List;
 import java.util.Set;
 
 import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IncrementalProjectBuilder;
+import org.eclipse.core.resources.ResourcesPlugin;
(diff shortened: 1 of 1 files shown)


ghost (approved):
Ok lets give it a try like this.

franzTobiasDLR (approved):

PhilMFischer:
Close #85

Issue #1228, opened 2024-10-30
VirSatServer does not have a default workspace


PhilMFischer:
The server in version 4.17.1 does not have a default workspace. This should be fixed in the product configuration.

PR #1230, opened 2024-10-30
Fixing server standard workspace

ghost:
Adjusted the product configuration to set the standard server workspace to "server".

So far no server workspace was defined and had to be set manually in the .ini file. Otherwise the serv er would start up with error messages regarding "no default workspace"


franzTobiasDLR (approved):

pchrszon-dlr (approved):

PhilMFischer:
Closes #1228

Issue #1196, opened 2024-08-12
Creation of SEIs is not taken anme into cosnideration and provides wrong type in response


dellerDLR:
**Describe the bug**
The query Parameter name is  not used by the server, when a new SEI is created.
Furthermore the type property is set incorrectly.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 7]

PR #1197, opened 2024-08-12
Task #1196 - Use the provided query parameter name when creating a SEI

dellerDLR:
Currently the post route of the SEi resource does not consider the name query parameter (eventhough its states in the OpenAPI spec).
Furthermore the type property in the response was wrongly set, which is also fixed.

PR #1237, opened 2024-11-12
Feature/1196 post request fix name param

franzTobiasDLR:
Re-done the changes from PR #1197 on new code base of 4.17.X

Currently the post route of the SEi resource does not consider the name query parameter (eventhough its states in the OpenAPI spec).
Furthermore the type property in the response was wrongly set, which is also fixed.


pchrszon-dlr (approved):

dellerDLR (approved):

PhilMFischer:
Closes #1196

PR #1251, opened 2024-12-20
Feature/task 1250 improve visual representation of state machine diagrams

### LINKED ISSUE #1250: Improve visual representation of state-machine diagrams
The state-machine diagram should be changed as follows:

- use more neutral colors
- draw states using rectangles instead of ellipses

ReyTamDLR:


pchrszon-dlr (approved):
Looking good!

PR #1253, opened 2025-02-03
Feature/task 101 rectangular lines by default in graphiti

### LINKED ISSUE #101: Rectangular lines by default in graphiti
Make a function that given a Graphiti Connection manipulates the Connection such that it is not visually a direct connection, but a rectangular one.

image

ReyTamDLR:
* Issue: #101 Rectangular Lines by Default in Graphiti

* Already accomplished To-Do:
	[X] Rectangular Lines as Default Transition Connection is now possible
* Accomplished in this commit:
	[X] Solved the issue where the anchor points and lines are offset when the distance 
	    between states are too small. Now it is possible for rectangular lines to be made
	    in between two states without overlapping, regardless of the states' position on
	    the diagram.


pchrszon-dlr (approved):
Great, looks good!

dellerDLR (approved):

